PR #21, opened 2026-01-28
Add CI for MSVC Build

playmer:
Hi there, neat project! I started taking a peek at some things, but figured I could get CI set up. This is mostly just creating a simple build.yml, but there's a couple of things to mention that may need to be changed.

- I've added an option to the CMake, by default, not compile the runtime. It seems like there's an assumption you're currently working on a recomp. If you're not, you don't have some of the expected files to compile the runtime, which causes some issues when building/running from VS. Probably what should happen is committing minimal sample versions of these files to the repo, allowing the runtime to be compiled as well. I can also switch the option to compile the runtime by default, and for now just disable it in the CI.
- The tests currently aren't all passing, but they are running in CI, but ignored. Once they pass, the `continue-on-failure` line in the yml should be removed.
- Obviously more platforms/compilers could be added here, but given the README mentions you're only testing on MSVC, I figured I'd start with that.


ran-j:
Nice I was thinking about add this.

ran-j:
Wait I didn't know that 2 test are failing

PR #34, opened 2026-01-30
Made member methods const

xms0g:
- Made some member methods const
- Used range-based loop instead of C-style one.
- Made one-variable constructor explicit
- forward declarations

### DIFF
--- a/ps2xAnalyzer/src/elf_analyzer.cpp
+++ b/ps2xAnalyzer/src/elf_analyzer.cpp
@@ -781,7 +781,8 @@ namespace ps2recomp
         }
     }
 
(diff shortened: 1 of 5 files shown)


ps2xAnalyzer/src/elf_analyzer.cpp
```diff
@@ -901,8 +900,7 @@ namespace ps2recomp
         }
     }
 
-    void ElfAnalyzer::analyzeRegisterUsage()
-    {
+    void ElfAnalyzer::analyzeRegisterUsage() const {
```
ran-j: Please keep Allman style.

ran-j:
@xms0g thanks for these changes

PR #45, opened 2026-02-05
Improve GCC/Clang portability and header cleanliness

Whoneon:
This PR focuses on portability and build reliability across GCC/Clang while avoiding risky and huge changes. It makes several headers self‑contained by adding missing `<cstdint> ` includes for fixed‑width integer types, in order to prevent order‑dependent build failures. The syscall `mutex` is moved to a single translation unit to avoid per‑TU instances and ensure consistent locking behavior. The runtime target now enables `-msse4.1` on GCC/Clang, aligning with the projects documented SSE4 requirements and fixing compile errors. 
Finally, a duplicate `FetchContent_Declare(libdwarf)` entry is removed for clarity.

Wrapping it, here is the changes:

- Header cleanliness for fixed‑width types `(<cstdint>)` in core headers.
- One global `g_sys_fd_mutex` in `ps2_syscalls.cpp` to avoid multiple static instances.
- Explicit `-msse4.1` on GCC/Clang for SSE4.1 intrinsics.
- Cleanup of duplicate CMake `FetchContent_Declare` for `libdwarf`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,3 +44,49 @@ jobs:
             build/ps2xAnalyzer/Release/*.exe
             build/ps2xRecomp/Release/*.exe
             build/ps2xTest/Release/*.exe
+
+  linux-gcc-clang:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - compiler: gcc
+            cc: gcc
+            cxx: g++
+          - compiler: clang
+            cc: clang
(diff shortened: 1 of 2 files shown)


Whoneon:
@emartinez-dev did this fix the CI as well? has this been tested again?

Whoneon:
If needed, I can provide a new PR after the weekend. Thanks everyone!

PR #71, opened 2026-02-20
3-operand MULT/MULTU &  delay slot exception handling

aslanhudajev:
This commit: `6821c69eb5674807569416cbac236510491f41a4 ` fixes a lot of runtime errors.
The MULT/MULTU fix is the most impactful I would say. Not super sure about the delay slot fix, it did however fix many issues in the game I am trying to get to run.

**Fix R5900 3-operand MULT/MULTU:** 
Standard MIPS MULT rs, rt only writes HI/LO. The R5900 extends this with a 3-operand form MULT rd, rs, rt that also writes the low 32 bits of the result into rd. The code generator was omitting the rd write, leaving the destination register stale. Any game code that reads rd after a multiply (instead of using MFLO) would get garbage, causing infinite loops and incorrect computations. This was found in Gauntlet: Dark Legacy's InitAtreeSeqs where the loop counter was computed via 3-operand MULT and never advanced.

**Fix delay slot exception handling:** 
When a MIPS exception occurs on a delay slot instruction, the CPU sets the BD (Branch Delay) bit in the Cause register and EPC points to the branch instruction, not the delay slot. The runtime's exception handler needs to know it was in a delay slot to correctly re-execute the branch after returning from the exception. The code generator now emits ctx->in_delay_slot = 1 / = 0 around delay slot instruction code, allowing raiseCop0Exception to set EPC and BD correctly and preserve the branch target.

### DIFF
--- a/ps2xRecomp/include/ps2recomp/ps2_recompiler.h
+++ b/ps2xRecomp/include/ps2recomp/ps2_recompiler.h
@@ -36,6 +36,9 @@ namespace ps2recomp
             std::vector<Function> &functions,
             std::unordered_map<uint32_t, std::vector<Instruction>> &decodedFunctions,
             const std::vector<Section> &sections);
+        static size_t ResliceEntryFunctions(
+            std::vector<Function> &functions,
+            std::unordered_map<uint32_t, std::vector<Instruction>> &decodedFunctions);
 
     private:
         ConfigManager m_configManager;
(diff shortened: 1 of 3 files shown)


.cursor/data/funDefsZDataError.md
```diff
@@ -0,0 +1,6525 @@
+
```
ran-j: ok you have some files like this one that is specific for you project  and can't be merge  like this one.

ran-j:
Also the build are fail

PR #108, opened 2026-04-06
elf_parser: clamp overlapping function boundaries from external sources

jlsandri:
## Problem

When function boundaries come from an external source (e.g. a Ghidra auto-analysis export), the reported ranges can overlap: a parent function's length is reported as including its nested sub-functions. A concrete example seen in the wild is a 52-byte function reported as 456 bytes because Ghidra's auto-analysis folded nested `sub_xxx` children into the parent's size.

When the recompiler imports these, `extractFunctions()` happily keeps both the oversized parent and the nested children. The parent then gets code-generated with all the sub-function bodies inlined, and the per-call PC-mismatch safety checks inside the inlined sub-calls cascade early-returns up the entire call chain at runtime.

## Fix

After merging all function sources in `extractFunctions()`, walk the sorted function list and clamp each function's end to the start of the next function. An exemption is carved out when the gap is `< 16` bytes so that the legitimate "entry-slice" pattern (multiple entry points into a single logical function, spaced a few instructions apart) is preserved.

## Scope

- One file touched: `ps2xRecomp/src/lib/elf_parser.cpp`
- +21 / -0
- No behavioural change for well-formed function maps — the clamp only fires when the imported data is actually inconsistent.

## Rationale

This is purely defensive. The recompiler already expects non-overlapping function ranges; this commit enforces that invariant at import time instead of letting the downstream codegen produce wrong output silently.

### DIFF
--- a/ps2xRecomp/src/lib/elf_parser.cpp
+++ b/ps2xRecomp/src/lib/elf_parser.cpp
@@ -797,6 +797,27 @@ namespace ps2recomp
             func.end = (nextStart > func.start) ? nextStart : (func.start + 4);
         }
 
+        // Clamp overlapping functions.  Ghidra sometimes reports parent functions
+        // whose range includes nested sub-functions (e.g. 0x1046B8 spanning
+        // 456 bytes when it should be 52).  Without this clamp the recompiler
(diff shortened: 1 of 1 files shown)


jlsandri:

Issue #86, opened 2026-02-26
Recompiler: incorrect codegen for branch-always thunks + fragment misdecode with mixed VU0/MMI


Boodax78:
Two independent codegen issues encountered while bringing up an ELF with entry discovery enabled. Both are reproducible.

Bug 1: b target thunks do not execute target function
When a function consists only of a branch-always (b target, no link), the recompiler emits:
ctx->pc = 0xXXXXXXXXu;
return;

Example:
0x11A548: 10000012    b 0x11A594

Generated C++:
ctx->pc = 0x11A594u;
return;

Problem:
The original jal at the call site already set $ra.
The branch-only thunk should transfer control to the target while preserving $ra.
The target function must execute normally and return via its own jr $ra.

Instead, the target function never runs. The caller resumes with an unexpected ctx->pc, breaking dispatcher-style return contracts.
Expected behavior: the thunk should effectively behave as:
FUN_0011a594_0x11a594(rdram, ctx, runtime);

Manual workaround: registerFunction(thunk_addr, target_function).
This appears to be a base codegen issue for branch-always instructions, independent of entry discovery.

Bug 2: Mixed VU0 + MMI fragments misdecoded and over-sliced
Fragments discovered through the entry discovery pass that originate inside a parent function and contain both VU0 (vmove, vsub, vrinit, vrxor) and MMI instructions (pextlw, pextuw, pcpyld) are decoded incorrectly.

Observed behavior:
Instructions are misidentified (e.g. decoded as MFC1 or unrelated opcodes).
Generated fragment is not functional as generated.
Fragment boundary detection overflows into the next function in the binary.

This only occurs for fragments created by the entry discovery pass. Top-level functions decoded normally do not exhibit this issue.

These appear to be two separate root causes, bug 1 is in base codegen for branch-always instructions, bug 2 is in decode and boundary logic for mixed VU0/MMI fragment slicing.

Tested on commit 8d1f1c5

Thanks

Boodax78:
One additional issue observed after commit 669114f 

A large number of recompiled functions are truncated mid-body, the generated C++ ends before the function's declared address range is complete, with no `return` and no `jumpTarget`. The last emitted C++ instruction is mid-function (e.g. `addiu`, `sd`, `jal`) and the remainder of the MIPS body is silently dropped.

Scope: 4360 affected functions out of ~10 700 total (~41%). Breakdown:
[...]

PR #109, opened 2026-04-06
codegen: safety fallback + entry-slice boundary fix for truncated functions (#86)

jlsandri:
Addresses #86. Two small, independent fixes for functions whose last instruction is not `jr \$ra` — either because the source really lacks a tail return, or because the imported function boundary is wrong and the real `jr \$ra` lives past the reported end.

## Fix 1 — Implicit safety fallback at function end (`code_generator.cpp`)

Emit an implicit
```cpp
ctx->pc = GPR_U32(ctx, 31);
return;
```
at the end of every generated function body.

- For well-formed functions this is unreachable dead code (the preceding `jr \$ra` already returned).
- For functions with incorrect boundaries, it returns cleanly via `\$ra` instead of leaving `ctx->pc` at whatever value the last fall-through instruction set, which previously caused cascading wrong returns up the call chain.

## Fix 2 — Skip boundary starts inside a containing parent (`ps2_recompiler.cpp`)

When reslicing entry functions, skip any boundary start that falls *inside* the range of the containing parent function.

Ghidra-style `sub_xxx` children inside a parent were creating "boundary" entries that truncated sibling entry slices before their `jr \$ra`, producing half-functions that would fall off the end at runtime.

## Scope

- `ps2xRecomp/src/lib/code_generator.cpp`: +6 / -0
- `ps2xRecomp/src/lib/ps2_recompiler.cpp`: +13 / -1
- No change in behaviour for correct inputs.

### DIFF
--- a/ps2xRecomp/src/lib/code_generator.cpp
+++ b/ps2xRecomp/src/lib/code_generator.cpp
@@ -1003,6 +1003,12 @@ namespace ps2recomp
             }
         }
 
+        // Safety fallback: if execution reaches here, the function boundary
+        // was incomplete (missing jr $ra). Return via $ra to prevent ctx->pc
+        // from being left in a bad state, which would cascade returns up the
+        // entire call chain. For correctly-terminated functions this is
+        // unreachable dead code after the jr $ra return block.
+        ss << "    ctx->pc = GPR_U32(ctx, 31); return;\n";
         ss << "}\n";
         return ss.str();
     }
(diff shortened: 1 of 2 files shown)


jlsandri:

ran-j:
Hey @jlsandri I didnt saw this PR sorry.

By looking ate issue #86 it seems to still has this bug, so if could please take a look again, my PR added a generic fallback to try fix that bug.

PR #110, opened 2026-04-06
codegen: remove dead else branch; narrow ps2_recompiled_functions.h include

jlsandri:
Small cleanup, no behavioural change.

## Changes

1. **Remove dead `else` branch in the call-emission path of `code_generator.cpp`.** The preceding `if` always returns or falls through in a way that makes the `else` unreachable.
2. **Stop including `ps2_recompiled_functions.h` in every generated translation unit.** The header is only consumed by `register_functions.cpp`, so including it in each generated TU was pure parse-time overhead with no functional effect.

## Scope

- `ps2xRecomp/src/lib/code_generator.cpp`: +13 / -21

## Rationale

Tidy-up discovered while working on larger codegen changes in follow-up PRs. Split out as its own PR to keep the review diff minimal and to unblock the follow-ups that touch the same file.

### DIFF
--- a/ps2xRecomp/src/lib/code_generator.cpp
+++ b/ps2xRecomp/src/lib/code_generator.cpp
@@ -315,7 +315,12 @@ namespace ps2recomp
 
                 if (!funcName.empty())
                 {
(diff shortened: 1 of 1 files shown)


jlsandri:

PR #111, opened 2026-04-06
codegen: emit lookupFunction dispatch for external jump-table targets

jlsandri:
## Problem

When a `[[jump_tables.table]]` entry in the TOML config has targets that live *outside* the enclosing function's boundary, the previous codegen silently filtered them out. Those external targets would then be unreachable from the recompiled jump, which at runtime manifests as a wrong branch destination (or a no-op, depending on the dispatch path).

## Fix

- Teach analysis to distinguish internal jump-table targets (which still use `goto`-to-label dispatch inside the function body) from external ones.
- Emit a runtime `lookupFunction(0x…)` dispatch call for the external targets, matching the fallback path already used for unresolved static `J`/`JAL` sites (per the README's "unresolved static J/JAL" behaviour).
- Add an `externalJumpTableTargets` field to `AnalysisResult` so the two categories can be separated during codegen instead of being decided in the emitter.

## Scope

- `ps2xRecomp/include/ps2recomp/code_generator.h`: +4 / -0
- `ps2xRecomp/src/lib/code_generator.cpp`: +46 / -9

## Rationale

The recompiler already has a principled fallback for unresolved static calls (`runtime->lookupFunction(...)`), so routing external jump-table targets through the same mechanism fits the existing model and makes the previously-filtered targets actually reachable at runtime.

### DIFF
--- a/ps2xRecomp/include/ps2recomp/code_generator.h
+++ b/ps2xRecomp/include/ps2recomp/code_generator.h
@@ -38,6 +38,10 @@ namespace ps2recomp
         struct AnalysisResult {
             std::unordered_set<uint32_t> entryPoints;
             std::unordered_map<uint32_t, std::vector<uint32_t>> jumpTableTargets;
+            // Targets from TOML-configured jump tables that are OUTSIDE the
+            // function boundary. Codegen emits lookupFunction calls for these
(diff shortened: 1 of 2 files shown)


jlsandri:

PR #112, opened 2026-04-06
codegen: fix VU0 dest-mask lane-bit ordering and VMR32 shuffle pattern typos

jlsandri:
## Problem

Two copy-paste typos in the VU0 macro codegen, both producing silently-wrong output for any instruction that uses a non-default dest-mask or the VMR32 rotate.

### 1. `dest_mask` lane-bit ordering inverted across ~36 VU0 emitters

Many emitters were passing the dest-mask bits to `_mm_set_epi32()` as
```cpp
(dest_mask & 0x8) ? -1 : 0, (dest_mask & 0x4) ? -1 : 0,
(dest_mask & 0x2) ? -1 : 0, (dest_mask & 0x1) ? -1 : 0
```
and the `VSQD` emitter additionally had
```cpp
{0x1, 0x2, 0x2, 0x1}
```
as a clearly-typo'd variant of the same pattern.

The correct ordering — matching the rest of the VU0 dest-mask code and the PS2 convention where bit 0 is the X lane — is
```cpp
(dest_mask & 0x1) ? -1 : 0, (dest_mask & 0x2) ? -1 : 0,
(dest_mask & 0x4) ? -1 : 0, (dest_mask & 0x8) ? -1 : 0
```

Because `_mm_set_epi32()` takes its arguments in **reverse lane order** (arg 0 → lane 3, arg 3 → lane 0), the inverted bit order caused the X/Y mask bits to gate the Z/W lanes and vice versa. Any masked write with a non-uniform dest_mask silently misrouted.

Affected emitters include: `VSQD`, `VADD_Field` / `VSUB_Field` / `VMUL_Field` / `VMADD_Field` / `VMSUB_Field`, `VMINI_Field` / `VMAX_Field`, `VFTOI*` / `VITOF*`, the `_bc` variants, the `_q` variants, the `_i` variants, and assorted `VMULA` / `ADDAq` sites. ~36 call sites in total.

### 2. `VU0_S2_VMR32` shuffle pattern

```cpp
// before
_MM_SHUFFLE(0, 0, 0, 1)
// after
_MM_SHUFFLE(0, 3, 2, 1)
```
`VMR32` is a 1-lane rotate, not a broadcast of lane 1. The old pattern produced `{lane1, lane1, lane1, lane1}` instead of the correct `{lane1, lane2, lane3, lane0}`.

## Scope

- +68 / -68 (almost all sites are pure re-orderings of an `_mm_set_epi32` call)
- Zero behavioural change for `dest_mask == 0xF` (the most common case); every other mask was silently wrong before.

## Rationale

Both classes of bug are pure copy-paste errors. The fix restores the stated invariant (bit 0 ↔ X lane) across every emitter, and corrects the VMR32 shuffle to perform the rotate the mnemonic actually describes.

### DIFF
--- a/ps2xRecomp/src/lib/code_generator.cpp
+++ b/ps2xRecomp/src/lib/code_generator.cpp
@@ -2359,14 +2359,14 @@ namespace ps2recomp
                                        "__m128i mask = _mm_set_epi32({}, {}, {}, {}); "
                                        "ctx->vu0_vf[{}] = _mm_blendv_ps(ctx->vu0_vf[{}], res, _mm_castsi128_ps(mask)); }}",
                                        inst.rd,
-                                       (dest_mask & 0x8) ? -1 : 0, (dest_mask & 0x4) ? -1 : 0,
-                                       (dest_mask & 0x2) ? -1 : 0, (dest_mask & 0x1) ? -1 : 0,
+                                       (dest_mask & 0x1) ? -1 : 0, (dest_mask & 0x2) ? -1 : 0,
+                                       (dest_mask & 0x4) ? -1 : 0, (dest_mask & 0x8) ? -1 : 0,
                                        inst.rt, inst.rt);
                 }
                 case VU0_S2_VMOVE:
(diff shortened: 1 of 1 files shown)


jlsandri:

PR #113, opened 2026-04-06
codegen: PS2-accurate divide-by-zero semantics for COP1 and VU0 FP + RSQRT operand fix

jlsandri:
## Problem

Five related floating-point codegen corrections for EE COP1 and VU0. All five are mis-emulations of documented PS2 hardware semantics.

### 1. `COP1_S_RSQRT` was using the wrong source operand

The previous emitter produced
```cpp
ctx->f[fd] = 1.0f / sqrtf(ctx->f[fs]);
```
which is the formula for reciprocal-sqrt of `fs`. The actual PS2 `rsqrt.s` operation is
```
fd = fs / sqrt(ft)
```
The emitter ignored `ft` entirely and used `fs` as the radicand. Fixed to use `ft` as the radicand and `fs` as the numerator. (This operand fix is also a prerequisite for the div-by-zero handling below, which rewrites the same line.)

### 2. `COP1_S_DIV` divide-by-zero

The PS2 EE does **not** produce IEEE infinity on divide-by-zero. It returns `0x7F7FFFFF` (the largest finite float) with the sign bit set to `sign(fs) XOR sign(ft)`. Previously the codegen emitted `copysignf(INFINITY, fs * 0.0f)` which produces IEEE ±∞.

### 3. `COP1_S_RSQRT` divide-by-zero

Same `0x7F7FFFFF` + sign semantics, with sign taken from `fs`. Also guards the `sqrt` input with `std::max(0.0f, ft)` to match the EE behaviour of ignoring the sign bit on the radicand instead of producing a NaN.

### 4. `VU0 vdiv` divide-by-zero

Same `0x7F7FFFFF` + sign semantics, writing to `vu0_q`. Previously returned `0.0f` on div-by-zero, which is visibly wrong for game code that relies on the max-float fallback.

### 5. `VU0 vrsqrt` divide-by-zero

Same `0x7F7FFFFF` semantics (unsigned, per the VU0 ISA). Previously returned `0.0f`.

## Implementation

All five use `std::bit_cast<float>(uint32_t)` for the bit-level construction of the max-float sentinel, and each site has an inline comment documenting the `0x7F7FFFFF` magic constant so it is self-explanatory.

## Scope

- +10 / -5

## Rationale

All five changes bring the generated code into line with documented PS2 floating-point semantics. The div-by-zero behaviour in particular is well-known to differ from IEEE and is frequently relied upon by EE code that intentionally divides by zero as a cheap max-float construction.

### DIFF
--- a/ps2xRecomp/src/lib/code_generator.cpp
+++ b/ps2xRecomp/src/lib/code_generator.cpp
@@ -1640,10 +1640,12 @@ namespace ps2recomp
             case COP1_S_MUL:
                 return fmt::format("ctx->f[{}] = FPU_MUL_S(ctx->f[{}], ctx->f[{}]);", fd, fs, ft);
             case COP1_S_DIV:
+                // PS2 EE div.s: division by zero returns max float (0x7F7FFFFF) with sign preserved, not IEEE infinity
                 return fmt::format("if (ctx->f[{}] == 0.0f) {{ ctx->fcr31 |= 0x100000; /* DZ flag */ "
-                                   "ctx->f[{}] = copysignf(INFINITY, ctx->f[{}] * 0.0f); }} "
+                                   "uint32_t __sign = (std::bit_cast<uint32_t>(ctx->f[{}]) ^ std::bit_cast<uint32_t>(ctx->f[{}])) & 0x80000000u; "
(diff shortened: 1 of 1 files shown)


jlsandri:

PR #115, opened 2026-04-06
codegen: handle continuation PCs and nested callee yields

jlsandri:
> **Stacked on #110** (codegen: remove dead else branch). Please merge #110 first — once it lands, this PR's diff shrinks to just the two commits below.

## Problem

The recompiler's "continuation PC" model assumed a 1:1 mapping between continuation targets and emitted entry functions, and that calls into functions with continuation PCs would always yield back at the single expected fallthrough address.

Two related failure modes:

1. **Continuation targets without a separate entry function.** When the analysis produced a continuation PC that did *not* have its own `entry_...` wrapper, the codegen path assumed the entry existed and produced a call that couldn't be dispatched.
2. **Nested callee yields.** When callee A yielded internally and re-entered itself at a continuation PC, and callee B was called from inside A, B's own continuation yield would bubble up as a `PC_MISMATCH` in A's caller — a false positive, because A was still mid-execution and just hadn't reached its outer fallthrough yet.

## Fix

Two commits:

1. **Handle continuation PCs without separate entry functions.** Teach `code_generator.cpp` / `ps2_recompiler.cpp` to dispatch continuation targets directly when no entry wrapper exists, instead of assuming the wrapper is always present. Adds a new bookkeeping field to the codegen header.

2. **Add a continuation re-entry loop for nested callee yields.** After a call, emit a `while` loop that keeps re-entering the callee while its yielded `ctx->pc` lies inside the callee's own range, only exiting to the caller when the real fallthrough PC is reached.

These go together — the re-entry loop depends on the first commit's bookkeeping to know where a continuation target lives.

## Scope

- `ps2xRecomp/include/ps2recomp/code_generator.h`: +3 / -0
- `ps2xRecomp/src/lib/code_generator.cpp`: +124 / -6
- `ps2xRecomp/src/lib/ps2_recompiler.cpp`: +41 / -17

## Rationale

Continuation PCs were already a supported concept in the recompiler; this PR just makes the handling robust to the two edge cases above, which were producing false `PC_MISMATCH` aborts in otherwise well-formed code.

### DIFF
--- a/ps2xRecomp/include/ps2recomp/code_generator.h
+++ b/ps2xRecomp/include/ps2recomp/code_generator.h
@@ -49,6 +49,7 @@ namespace ps2recomp
         void setBootstrapInfo(const BootstrapInfo &info);
         void setRelocationCallNames(const std::unordered_map<uint32_t, std::string> &callNames);
         void setConfiguredJumpTables(const std::vector<JumpTable> &jumpTables);
+        void setFunctionBounds(const std::vector<Function> &functions);
 
         AnalysisResult collectInternalBranchTargets(const Function &function,
                                                   const std::vector<Instruction> &instructions);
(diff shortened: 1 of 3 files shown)


jlsandri:

PR #116, opened 2026-04-06
codegen: add dest-mask support for VU0 VMOVE and VMR32

jlsandri:
> **Stacked on #112** (codegen: fix VU0 dest-mask lane-bit ordering and VMR32 shuffle pattern typos). Please merge #112 first — once it lands, this PR's diff shrinks to just the single commit below.

## Problem

`VU0_S2_VMOVE` and `VU0_S2_VMR32` were emitting unconditional moves/shuffles that ignored the instruction's `dest_mask` field. Any masked VMOVE/VMR32 therefore overwrote lanes the mask was meant to preserve, silently corrupting state for every masked write.

## Fix

Both emitters now:

1. Fast-path `dest_mask == 0xF` to the existing unconditional codegen (identical output for the common case — zero overhead).
2. For any other mask, emit an `_mm_blendv_ps` guarded by a per-lane selector built from the dest_mask bits, using the same `{0x1, 0x2, 0x4, 0x8}` lane-bit convention as the rest of the VU0 dest-mask emitters.

## Why this is stacked on #112

`VMR32`'s masked path emits a shuffle followed by a blend; the shuffle uses `_MM_SHUFFLE(0,3,2,1)`, which is the corrected form from #112. Without that prerequisite merged, the masked VMR32 would rotate to the wrong lane — the same typo #112 fixes in the unmasked path.

## Scope

- +25 / -2

## Rationale

Parity with the rest of the VU0 macro codegen. Every other vector instruction already respects `dest_mask`; `VMOVE`/`VMR32` were outliers.

### DIFF
--- a/ps2xRecomp/src/lib/code_generator.cpp
+++ b/ps2xRecomp/src/lib/code_generator.cpp
@@ -2364,9 +2364,32 @@ namespace ps2recomp
                                        inst.rt, inst.rt);
                 }
                 case VU0_S2_VMOVE:
-                    return fmt::format("ctx->vu0_vf[{}] = ctx->vu0_vf[{}];", inst.rt, inst.rd);
+                {
+                    uint8_t dest_mask = inst.vectorInfo.vectorField;
+                    if (dest_mask == 0xF) {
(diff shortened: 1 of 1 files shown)


jlsandri:

PR #117, opened 2026-04-06
codegen: force continuation on JAL/JALR resume-PC mismatch instead of aborting caller

jlsandri:
> **Stacked on #115** (codegen: handle continuation PCs and nested callee yields). Please merge #115 first — once it lands, this PR's diff shrinks to just the two commits below.

## Problem

After a `JAL`/`JALR` call + continuation re-entry loop, if `ctx->pc` didn't match the expected fallthrough address, the previous codegen logged `[PC_MISMATCH]` and aborted the caller with `return`.

That abort cascaded up the entire call stack whenever a nested callee legitimately left `ctx->pc` at an internal resume target — common with overlay functions, nested `JALR` chains, and any function that sets `ctx->pc` as part of its own resume-PC mechanism. The caller's stack frame and register context were still valid; the codegen was throwing them away on a false signal.

## Fix

Replace the hard `return` with:

1. A throttled `[JALR_FIXUP]` / `[JAL_FIXUP]` diagnostic — 5 warnings per call site, so the signal isn't lost.
2. `ctx->pc = fallthroughPc;` — force continuation at the caller's expected resume point.

Two commits:

1. **`JALR` resume-PC fixup** at the single indirect-call site (depends on #115 for the re-entry loop structure it edits).
2. **`JAL` resume-PC fixup** extending the same treatment to all three direct-call codegen paths for consistency:
   - Direct `JAL` via `findFunctionByAddress` (`called 0x%x` pattern)
   - Direct `JAL` via `lookupFunction` (`called 0x{literal}` pattern)
   - Reloc `JAL` (`called reloc` pattern)

## Why this is stacked on #115

The JALR hunk rewrites the `PC_MISMATCH` branch that #115 introduces as part of the re-entry loop. Without #115 merged first, there is no `PC_MISMATCH` to rewrite.

## Scope

- +21 / -7

## Rationale

The old behaviour was conservative — any resume-PC anomaly aborted the whole chain — and that's reasonable as a first line of defence. But in practice the only trigger is legitimate internal resume-PC use by the callee, which the caller should tolerate. Demoting the abort to a throttled warning + forced continuation keeps the diagnostic signal while eliminating the cascade-failure class entirely.

### DIFF
--- a/ps2xRecomp/include/ps2recomp/code_generator.h
+++ b/ps2xRecomp/include/ps2recomp/code_generator.h
@@ -49,6 +49,7 @@ namespace ps2recomp
         void setBootstrapInfo(const BootstrapInfo &info);
         void setRelocationCallNames(const std::unordered_map<uint32_t, std::string> &callNames);
         void setConfiguredJumpTables(const std::vector<JumpTable> &jumpTables);
+        void setFunctionBounds(const std::vector<Function> &functions);
 
         AnalysisResult collectInternalBranchTargets(const Function &function,
                                                   const std::vector<Instruction> &instructions);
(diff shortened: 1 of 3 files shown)


jlsandri:

Issue #127, opened 2026-06-03
Process gets killed because it's using too much memory


Sinan-Karakaya:
Hi. When using ps2_recomp on an ELF which contains ~120k entry points and ~7k owner functions, the process gets killed before it finishes because it uses too much memory.

Memory usage is increasing steadily during execution, and gets killed when usage goes at around 90%. I have 32GB of memory on my system, which seems reasonably sufficient for such a task.

OS: Ubuntu 26.04 LTS

I'm currently taking a look into it, see if the fix is trivial. But in any case, any insight is welcomed in case I'm doing something wrong :)

Sinan-Karakaya:
Edit: Tried on a VM with 64GB of memory + 16GB of swap space, it was still using 100% of both

ran-j:
thanks for report, can  you share more info like your toml file

Sinan-Karakaya:
Absolutely, here is my .toml file

```toml
# Auto-generated by ExportPS2Functions.java
#
# Classification policy (aligned with analyzer intent):
# - library/runtime names -> [general].stubs
# - system names -> [general].skip
# - others are left for recompilation

[general]
input = ".../SLUS_XXX.XX"
output = ".../out/output"
ghidra_output = ".../out/output.csv"
single_file_output = false
patch_syscalls = false
patch_cop0 = true
patch_cache = true
stubs = [
  "SetGsCrt@0x00116420",
  "_EnableIntc@0x00116560",
  "_DisableIntc@0x00116570",
  "_EnableDmac@0x00116580",
  "_DisableDmac@0x00116590",
  "CreateThread@0x00116620",
  "DeleteThread@0x00116630",
  "_StartThread@0x00116640",
  "ExitDeleteThread@0x00116660",
  "TerminateThread@0x00116670",
  "SleepThread@0x00116740",
  "WakeupThread@0x00116750",
  "_iWakeupThread@0x00116760",
  "SuspendThread@0x00116790",
  "EndOfHeap@0x00116800",
  "CreateSema@0x00116820",
  "DeleteSema@0x00116830",
  "SignalSema@0x00116840",
  "iSignalSema@0x00116850",
  "WaitSema@0x00116860",
  "PollSema@0x00116870",
  "iPollSema@0x00116880",
  "SetOsdConfigParam@0x001168C0",
  "GetOsdConfigParam@0x001168D0",
  "FlushCache@0x00116AA0",
  "sceSifStopDma@0x00116AF0",
[...]

PR #128, opened 2026-06-03
Reduce recompiler output memory usage, and added multi-threading to recomp process

Sinan-Karakaya:
## Summary

This reduces peak memory usage during recompilation/output generation and avoids pathological generated C++ growth for large retail ELFs.

Changes include:

- Add `general.low_memory_mode`.
- Add `general.output_worker_threads` with bounded auto worker selection.
- Stream combined output instead of buffering the entire generated file in memory.
- Bound individual/combined output generation workers.
- Avoid retaining Rabbitizer disassembly strings in low-memory mode.
- Preserve generated asm comments by producing disassembly text on demand while writing output.
- Avoid emitting huge repeated indirect-jump fallback switches for unresolved indirect branches.
- Register indirect fallback entry points as resumable entries instead.
- Clear decoded/generated buffers once output no longer needs them.
- Document the new config options in the README and example config.

## Why

Large stripped ELFs with many functions/internal entry points could cause the recompiler/output stage to retain or generate enormous buffers. In one test case, generated output included an 8+ GB `.cpp` file due repeated indirect-jump fallback switches.

We change how unresolved indirect jump fallback is emitted. Known/configured jump tables still generate direct local switches, but when an indirect jump target cannot be resolved, we no longer emit a large switch over every possible internal label at each indirect branch site. Instead, those fallback targets are registered as resumable entry points and rely on runtime dispatch to re-enter the owning function at the correct ctx->pc.

This avoids pathological multi-GB generated C++ files, but it may be worth watching in future testing. The main risk would be code paths that depend on an unresolved internal indirect jump staying inside the same generated C++ stack frame rather than returning through runtime dispatch. Areas with jump tables, callbacks, function pointers, script dispatch, battle logic, or menu state machines are good candidates for extra runtime validation.

## Validation

Tested with the  config and ELF mentionned in #127 :

```bash
cmake --build out/build --target ps2_recomp
/usr/bin/time -v out/build/ps2xRecomp/ps2_recomp out/config.toml
```

Observed results after the changes:
- Peak RSS: about 471 MB
- Output directory size with comments enabled: about 355 MB
- Recompiler completed successfully
- Generated .cpp files still include original address/raw/disassembly comments

Although behavior should remain the same, we could benefit from extensive testing when it comes to indirect jumps. If there are any small, test focused ELF for this I'd like someone to give it a try and compare the outputs of the source and destination branches recompilations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Main fields in `config.toml`:
 * `general.output`: generated C++ output folder.
 * `general.single_file_output`: one combined cpp or one file per function.
 * `general.low_memory_mode`: reduce peak output-generation memory by avoiding retained disassembly strings and forcing serial output generation. Generated instruction comments are still emitted; disassembly text is produced while writing each output file instead of being kept in memory.
-* `general.output_worker_threads`: number of output-generation workers. A positive value uses exactly that many workers. `0` uses `nproc - 1` when at least two hardware threads are available, otherwise serial output generation. `1` forces serial output generation.
+* `general.output_worker_threads`: number of output-generation workers (clamped to nproc * 2). A positive value uses exactly that many workers. `0` uses `nproc - 1` when at least two hardware threads are available, otherwise serial output generation. `1` forces serial output generation.
 * `general.patch_syscalls`: apply configured patches to `SYSCALL` instructions (`false` recommended).
 * `general.patch_cop0`: apply configured patches to COP0 instructions.
 * `general.patch_cache`: apply configured patches to CACHE instructions.
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ Main fields in `config.toml`:
 * `general.ghidra_output`: recommended function map CSV exported from Ghidra.
 * `general.output`: generated C++ output folder.
 * `general.single_file_output`: one combined cpp or one file per function.
+* `general.low_memory_mode`: reduce peak output-generation memory by avoiding retained disassembly strings and forcing serial output generation. Generated instruction comments are still emitted; disassembly text is produced while writing each output file instead of being kept in memory.
+* `general.output_worker_threads`: number of output-generation workers. A positive value uses exactly that many workers. `0` uses `nproc - 1` when at least two hardware threads are available, otherwise serial output generation. `1` forces serial output generation.
 * `general.patch_syscalls`: apply configured patches to `SYSCALL` instructions (`false` recommended).
 * `general.patch_cop0`: apply configured patches to COP0 instructions.
 * `general.patch_cache`: apply configured patches to CACHE instructions.
@@ -118,6 +120,8 @@ ghidra_output = ""
 output = "output/"
 
 single_file_output = true
+low_memory_mode = true
+output_worker_threads = 0
 patch_syscalls = false
 patch_cop0 = true
 patch_cache = true
--- a/ps2xRecomp/src/lib/code_generator.cpp
+++ b/ps2xRecomp/src/lib/code_generator.cpp
@@ -1,9 +1,9 @@
 #include "ps2recomp/code_generator.h"
 #include "ps2recomp/instructions.h"
 #include "ps2recomp/ps2_recompiler.h"
+#include "ps2recomp/r5900_decoder.h"
 #include "ps2recomp/types.h"
 #include "ps2_runtime_calls.h"
-#include "rabbitizer.h"
 #include <fmt/format.h>
 #include <sstream>
 #include <algorithm>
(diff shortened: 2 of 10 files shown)


Sinan-Karakaya:
As a side note, I am also planning to add a progress tracking if necessary once this PR gets merged (if it does).
Although that might not be needed, the recompilation took less than a minute on my system, whereas it used to crash after running for ~15mins.

ps2xRecomp/src/lib/code_generator.cpp
```diff
@@ -259,20 +295,14 @@ namespace ps2recomp
...
                 std::sort(sortedInternalTargets.begin(), sortedInternalTargets.end());
-            } else {
-                sortedInternalTargets.reserve(internalTargets.size());
-                for (uint32_t t : internalTargets)
-                {
-                    sortedInternalTargets.push_back(t);
-                }
-                std::sort(sortedInternalTargets.begin(), sortedInternalTargets.end());
             }
         }
```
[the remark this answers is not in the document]
Sinan-Karakaya: I did not restore the old `internalTargets -> sortedInternalTargets` fallback because that is the path that generated the multi-GB repeated switches.

I added explicit comments documenting the intended behavior:
- direct local switches are emitted only for resolved jump tables
- unresolved indirect fallback targets are kept as labels and registered as resumable entries for runtime dispatch.

ran-j:
Nice work I will review later.

ps2xRecomp/src/lib/code_generator.cpp
```diff
@@ -1010,6 +1044,9 @@ namespace ps2recomp
                     if (addr >= function.start && addr < function.end)
                     {
                         result.entryPoints.insert(addr);
+                        // Keep labels and runtime registration for unresolved indirect jumps
+                        // without emitting a local switch over every possible target.
+                        result.indirectFallbackEntryPoints.insert(addr);
```
ran-j, line 1022: Can we add coverage for unresolved `JALR` targets here, or can you explain why we do not need the same resumable-entry fallback as `JR`? 

The old local-switch fallback was removed for both `JR` and `JALR`, but `indirectFallbackEntryPoints` is only populated through `needsJrFallback`, and that excludes `SPECIAL_JALR`. 

So If a `JALR` target can be an internal label that is not discovered as a jump table/resume entry, generated code will call `runtime->lookupFunction(jumpTarget)` instead of re-entering the owning function at that label.

This can cause not found address on runtime

ran-j, line 1022: I think it needs a focused regression test for this new semantic change.

I think the pr should cover an unresolved `JR` to an internal label and verify that the target is registered as a resumable entry for the owning function, while resolved jump tables still emit local switch cases.

README.md
```diff
@@ -92,6 +92,8 @@ Main fields in `config.toml`:
 * `general.ghidra_output`: recommended function map CSV exported from Ghidra.
 * `general.output`: generated C++ output folder.
 * `general.single_file_output`: one combined cpp or one file per function.
+* `general.low_memory_mode`: reduce peak output-generation memory by avoiding retained disassembly strings and forcing serial output generation. Generated instruction comments are still emitted; disassembly text is produced while writing each output file instead of being kept in memory.
+* `general.output_worker_threads`: number of output-generation workers. A positive value uses exactly that many workers. `0` uses `nproc - 1` when at least two hardware threads are available, otherwise serial output generation. `1` forces serial output generation.
```
ran-j: This says a positive `output_worker_threads` value uses exactly that many workers, but the config loader clamps it to `hardware_concurrency() * 2`.

ps2xRecomp/src/lib/code_generator.cpp
```diff
@@ -109,6 +111,30 @@ namespace ps2recomp
         return kKeywords.contains(name);
     }
 
+    static std::string getInstructionDisassembly(const Instruction &inst)
```
ran-j: Can we move this to R5900Decoder like `R5900Decoder::decodeInstruction(raw, address)`

The ideia is to keep Rabbitizer/disassembly formatting in the decoder layer instead of adding it to `CodeGenerator` .

ps2xRecomp/src/lib/code_generator.cpp
```diff
@@ -259,20 +295,18 @@ namespace ps2recomp
 
         std::vector<uint32_t> sortedInternalTargets;
         if (branchInst.opcode == OPCODE_SPECIAL &&
-            (branchInst.function == SPECIAL_JR || branchInst.function == SPECIAL_JALR) &&
+            ((branchInst.function == SPECIAL_JR && branchInst.rs != 31) ||
+             branchInst.function == SPECIAL_JALR) &&
             !internalTargets.empty())
         {
+            // Only emit local indirect-jump switches for jump tables we actually resolved.
```
ran-j, line 278: This comment says unresolved fallback targets are registered as resumable entries, but that seems to only happen for unresolved `JR`, not `JALR`. This condition includes `SPECIAL_JALR`, yet the fallback registration path below explicitly excludes `SPECIAL_JALR` from setting `needsJrFallback`. Is the intent that unresolved `JALR` can never target internal labels?

ran-j (changes requested):
This comment says unresolved fallback targets are registered as resumable entries, but that seems to only happen for unresolved `JR`, not `JALR`. This condition includes `SPECIAL_JALR`, yet the fallback registration path below explicitly excludes `SPECIAL_JALR` from setting `needsJrFallback`. Is the intent that unresolved `JALR` can never target internal labels? If so, please document that separately; otherwise this may drop the old local-switch behavior without adding equivalent resumable registrations.

ps2xRecomp/src/lib/code_generator.cpp
Sinan-Karakaya: Changed it, now unresolved JALR uses the same indirect fallback registration path as unresolved non-RA JR, so internal dynamic targets can re-enter the owning function instead of risking `runtime->lookupFunction(jumpTarget)` failing.

I added regression tests covering the following:
- unresolved JR fallback entries
- unresolved JALR fallback entries
- unresolved fallback avoids broad local switches
- resolved configured jump tables still emit local switch cases

ps2xRecomp/src/lib/code_generator.cpp
Sinan-Karakaya: No, the intent is not that unresolved JALR can never target internal labels. That was a mistake on my side. I changed the fallback flag from JR-only to generic indirect fallback, so unresolved JR and JALR both register resumable entry points for the owning function. I also updated the comment and added regression coverage for unresolved JALR internal fallback targets.

ran-j (approved):

ran-j:
nice work

PR #129, opened 2026-06-04
refactor: migrate GUI from SDL2/OpenGL to raylib and rlImGui for improved platform abstraction.

BlackLineInteractive:


ps2xStudio/CMakeLists.txt
```diff
@@ -37,42 +37,37 @@ FetchContent_Declare(
 )
 
 FetchContent_Declare(
-    SDL2
-    GIT_REPOSITORY https://github.com/libsdl-org/SDL.git
-    GIT_TAG        release-2.32.10
+    rlImGui
+    GIT_REPOSITORY https://github.com/raylib-extras/rlImGui.git
+    GIT_TAG        main
```
ran-j, line 42: can we use a tag here to prevent break on updates since raylib-extras depends on especifc verions of raylib

ran-j (changes requested):

PR #132, opened 2026-06-13
GS rasterizer rework

tealalchemist:
This PR reimplements a lot of the swizzling for gs memory.
Additionally, it provides some speed improvements by reducing a lot of the swizzling math to a lookup. Note that you need to compile with optimization to see this performance increase.

There is additionally a small `types.h` header I hope to work into the rest of the runtime at some point

**New texture formats:**
P8H
P4HL
P4HH
Z32
Z16
Z24
Z16S

**Improved over the previous code:**
CT16
CT16S

**Other improvements**:
GS memory wrapping (memory wraps at 4mb)

Test program: https://github.com/tealalchemist/gstests/tree/main/texture
Tested against emulator

output before this PR (provided by ranj, 24 bit is actually a bug I need to correct in the test program later)
image

output with this PR:
image

Emulator comparison:
image

The emulator is a modified version of DobieStation (used because it's pretty accurate and I find it easier to read than PCSX2)

**TODO:**
The clut on the GS is cached, this can be implemented to our benefit to reduce the clut swizzling calculations
Some of the code is repetitive. I hope to rework some of the transfer code to a generic `Unpack` function
There is currently no depth handling
Local to local and local to host transfer rework
Remove old swizzling code
Various other smaller improvements I think need to be made

Thank you to ranj for the help with getting the test program working properly

### DIFF
--- a/ps2xRuntime/CMakeLists.txt
+++ b/ps2xRuntime/CMakeLists.txt
@@ -167,6 +167,7 @@ add_library(ps2_runtime STATIC
     src/lib/ps2_audio.cpp
     src/lib/ps2_audio_vag.cpp
     src/lib/ps2_gs_gpu.cpp
+    src/lib/ps2_gs_memory.cpp
     src/lib/ps2_gs_rasterizer.cpp
     src/lib/ps2_iop.cpp
     src/lib/ps2_iop_audio.cpp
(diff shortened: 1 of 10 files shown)


tealalchemist:
we have z (note this test program relies on some rework of local to local transfers I haven't committed yet)

image

PR #136, opened 2026-06-19
fix: semaphore syscalls return sid on success instead of KE_OK

smmathews:
## What's wrong

On the real PS2 EE BIOS, the semaphore syscalls return the semaphore ID on success, not zero. Our implementations of \`PollSema\`, \`WaitSema\`, \`SignalSema\`, and \`DeleteSema\` were all returning \`KE_OK\` (0) on success instead.

## Why it matters

DQ8 is what surfaced this. Its init code at \`0x12A670\` polls a mutex semaphore and compares the result against the semaphore ID (\`bne \$v1, \$v0\`). When \`PollSema\` returns \`KE_OK\` instead of the ID, that comparison never matches, so the game spins forever at \`PC 0x164978\` and never finishes initializing. With the correct return value it proceeds normally.

This isn't DQ8-specific. The standard EE SDK semaphore pattern relies on the success return being the sema ID, so any game built against it can hit the same problem. \`CreateSema\` already returned the ID on success; the other four operations were just inconsistent with it.

## What changed

Five edits in \`ps2xRuntime/src/lib/Kernel/Syscalls/Sync.cpp\`:

- \`PollSema\`, \`SignalSema\`, \`DeleteSema\` now return \`sid\` on success instead of \`KE_OK\`.
- \`WaitSema\` seeds \`ret = sid\` instead of \`ret = 0\`. The error paths (\`KE_WAIT_DELETE\`, \`KE_RELEASE_WAIT\`) are negative and still override \`ret\`.
- \`WaitSema\`'s count-decrement guard changes from \`ret == 0\` to \`ret == sid\`. This one is required, not cosmetic: once the success value is a positive \`sid\`, the old \`== 0\` check would never fire, so the semaphore count would never be decremented and token accounting would break for every subsequent wait. Using \`ret == sid\` rather than \`ret >= 0\` also makes the success condition explicit — \`ret\` is seeded to \`sid\` and only ever overwritten with negative error codes, so the guard precisely expresses "this wait acquired this semaphore."

\`ReferSemaStatus\` is intentionally left returning \`KE_OK\` — it's a query, not an acquire/release operation, and the convention only applies to the state-changing ops.

## References

Public docs are thorough for \`CreateSema\` but sparse on the success return values of the other four. The behavior is inferred from game code plus two sources that are consistent with it:

- DQ8's \`bne \$v1, \$v0\` at \`0x12A670\`, comparing the \`PollSema\` return directly against the sema ID, is the concrete proof — the BIOS has to be returning \`sid\` for that branch to make sense.
- [ps2tek EE syscall reference](https://israpps.github.io/ps2tek/PS2/BIOS/EE_Syscalls.html) documents \`CreateSema\` returning the semaphore ID on success. The four state-changing ops follow the same convention.
- [ps2sdk \`kernel.h\`](https://github.com/ps2dev/ps2sdk/blob/master/ee/kernel/include/kernel.h) declares all of these with an \`s32\` return type, consistent with returning the ID (negative on error, positive ID on success).

## Tests

Existing semaphore tests updated to expect \`sid\` on success paths. New cases added:

- Blocked \`WaitSema\` woken by \`SignalSema\` returns \`sid\` (the DQ8 scenario directly).
- \`WaitSema\` force-released via \`ReleaseWaitThread\` returns \`KE_RELEASE_WAIT\` and does not consume a token (guards the \`ret == sid\` change).
- Count-decrement assertion after a successful wait confirms the guard still decrements correctly.

### DIFF
--- a/ps2xRuntime/src/lib/Kernel/Syscalls/Sync.cpp
+++ b/ps2xRuntime/src/lib/Kernel/Syscalls/Sync.cpp
@@ -218,7 +218,8 @@ namespace ps2_syscalls
         }
         sema->cv.notify_all();
 
-        setReturnS32(ctx, KE_OK);
+        // PS2 EE BIOS returns sid on success.
+        setReturnS32(ctx, sid);
     }
 
     void iDeleteSema(uint8_t *rdram, R5900Context *ctx, PS2Runtime *runtime)
@@ -236,7 +237,8 @@ namespace ps2_syscalls
             return;
         }
 
-        int ret = KE_OK;
+        // PS2 EE BIOS returns sid on success; KE_SEMA_OVF overrides on overflow.
+        int ret = sid;
         int beforeCount = 0;
         int afterCount = 0;
         {
@@ -286,7 +288,7 @@ namespace ps2_syscalls
         auto info = ensureCurrentThreadInfo(ctx);
         throwIfTerminated(info);
         std::unique_lock<std::mutex> lock(sema->m);
-        int ret = 0;
+        int ret = sid;  // PS2 EE BIOS returns sid on success.
 
         if (sema->count == 0)
         {
@@ -318,7 +320,7 @@ namespace ps2_syscalls
                               {
                                   bool forced = info ? info->forceRelease.load() : false;
                                   bool terminated = info ? info->terminated.load() : false;
-                                  return sema->count > 0 || sema->deleted || forced || terminated; //
+                                  return sema->count > 0 || sema->deleted || forced || terminated;
                               });
             }
             sema->waiters--;
@@ -346,7 +348,7 @@ namespace ps2_syscalls
             }
         }
 
-        if (ret == 0 && sema->count > 0)
+        if (ret >= 0 && sema->count > 0)
         {
             sema->count--;
         }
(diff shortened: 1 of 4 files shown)


ran-j:
Hey, I will review this one later.

Also why you close your other PR ?

smmathews:
I'm going to take a different approach on the other PR, I think it was a little too conservative.

Also, I'm working on the PR test fix. I appreciate the review, thanks.

ran-j:
Nice I see  https://github.com/ps2dev/ps2sdk/blob/master/ee/kernel/include/kernel.h#L423 declare `s32 SignalSema(s32 sema_id);` and macros just returns what BIOS put on $v0/$2

ps2xRuntime/src/lib/Kernel/Syscalls/Sync.cpp
```diff
@@ -346,7 +348,7 @@ namespace ps2_syscalls
             }
         }
 
-        if (ret == 0 && sema->count > 0)
+        if (ret >= 0 && sema->count > 0)
```
ran-j: I think is better to check `ret == sid`

ran-j (changes requested):

ps2xRuntime/src/lib/Kernel/Syscalls/Sync.cpp
smmathews: agreed. done.

Issue #142, opened 2026-06-30
Runtime: No explicit ACC register in FPU


jeff-shulkin:
Currently the code generator uses f[32] as the COP1/FPU ACC register without explicit documentation. To eliminate potentially dangerous behavior and make the code a little more self-documenting, we should add an explicit field for the FPU ACC register and add/alter appropriate macros and codegen cases to accomodate it.

PR #143, opened 2026-06-30
runtime: Added explicit FPU ACC register to code generator

jeff-shulkin:
Addresses #142. This PR adds a macro for setting the FPU ACC register, as well as altering other codegen cases to explicitly reference new field ``ctx->f_acc`` instead of ``ctx->f[31]``.


ran-j:
Perfect, can you please add at least one test for this?

PR #148, opened 2026-07-07
GS PRIM must honor PRMODECONT (attribute-bit clobber)

smmathews:
## Summary

`GS::writeRegister`'s `case GS_REG_PRIM` unconditionally overwrote all eight PRIM attribute bits (IIP/TME/FGE/ABE/AA1/FST/CTXT/FIX, bits 3-10) on every PRIM write, ignoring `PRMODECONT`. Per GS User's Manual PRMODECONT (AC) semantics, those attributes come from PRIM only when `PRMODECONT=1`; when `PRMODECONT=0` they are held from the last `PRMODE` write, and PRIM changes only the primitive TYPE (bits 0-2).

The `case GS_REG_PRMODE` handler was already correctly gated on `!m_prmodecont`; this change makes the PRIM case symmetric by wrapping its attribute-bit assignments in `if (m_prmodecont)`. `m_prim.type` and the vertex-queue reset remain unconditional.

Real-world impact: the common composite/present pattern `PRMODECONT=0` + `PRMODE(TME=1)` + bare `PRIM=SPRITE` per draw had its TME silently reset to 0 on every draw, so fullscreen textured blits rendered as flat untextured fills (observed as DQ8's present pass to `FRAME.FBP=0x70` showing only a gray clear).

## Changes

- `ps2xRuntime/src/lib/ps2_gs_gpu.cpp`: gate the eight PRIM attribute-bit writes on `m_prmodecont`.
- `ps2xTest/src/ps2_gs_tests.cpp`: two regression tests in the PS2GS suite:
  - **State-level** — `PRMODECONT=0` + `PRMODE(TME)` + bare `PRIM=SPRITE` must leave TME set (observed via `getDebugSnapshot().prim.tme`); then `PRMODECONT=1` + the same PRIM write must clear it.
  - **Render-level** — a sprite copy whose TME|FST come entirely from PRMODE (PRMODECONT=0, PRIM attribute bits 0) must sample its source texture (TEX0 uses DECAL so the result cannot alias the flat RGBAQ color); asserts the destination pixel equals the source texel and not the flat fill.

## Verification

- Full build succeeds (Debug, GCC 16; local-only `-include cstdint` workaround for vendored ELFIO header — not committed).
- Full `ps2x_tests` suite: **298/298 passed**, including both new cases.
- Regression-guard check: with the fix temporarily reverted, exactly the two new tests fail (296/298); restored, back to 298/298.

## Notes / limitations

- The render-level test seeds the source framebuffer via the existing `writeReferenceFramePSMCT32Pixel` test helper instead of drawing an untextured sprite pass — same coverage of the PRIM/PRMODE interaction under test, matching the established style of neighboring GS tests.
- No diagnostics logging was added here; that is out of scope for this change.

### DIFF
--- a/ps2xRuntime/src/lib/ps2_gs_gpu.cpp
+++ b/ps2xRuntime/src/lib/ps2_gs_gpu.cpp
@@ -1572,15 +1572,22 @@ void GS::writeRegister(uint8_t regAddr, uint64_t value)
     {
     case GS_REG_PRIM:
     {
+        // GS User's Manual PRMODECONT (AC) semantics: TYPE (bits 0-2) always
+        // comes from PRIM, but the attribute bits (bits 3-10) come from PRIM
+        // only when PRMODECONT=1; when PRMODECONT=0 they are held from the last
+        // PRMODE write (mirrors the gate already present in case GS_REG_PRMODE).
         m_prim.type = static_cast<GSPrimType>(value & 0x7);
-        m_prim.iip = ((value >> 3) & 1) != 0;
-        m_prim.tme = ((value >> 4) & 1) != 0;
-        m_prim.fge = ((value >> 5) & 1) != 0;
-        m_prim.abe = ((value >> 6) & 1) != 0;
-        m_prim.aa1 = ((value >> 7) & 1) != 0;
-        m_prim.fst = ((value >> 8) & 1) != 0;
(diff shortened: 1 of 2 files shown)


smmathews:
@tealalchemist is adding to their PR, no need for separate PR

PR #150, opened 2026-07-07
feat(recomp): discover packed jal-only entries via cross-unit manifests and thread-entry scan

smmathews:
## Problem

A callee reached only by a `jal`/`j` immediate from a caller outside its own byte range is
disassembled as unreachable tail code inside whichever function unit contains it, with no callable
entry of its own. At runtime `hasFunction(target)` fails and `dispatchGuestBranch` falls
into the missing-function policy; the default, `ContinueToTarget`, moves `pc` to the target without
executing the callee. `discoverAdditionalEntryPoints()` already recovers same-invocation
cross-function targets. Two forms are not recovered: callers in a separately recompiled unit or
overlay, and thread entry pointers stored in a static `ThreadParam` struct.

## Fix

Two discovery paths, both registering through the existing `m_resumeEntryTargetsByOwner` mapping.
No runtime or code-generator behavior changes.

**Cross-unit manifests.** Each invocation writes `external_call_targets.txt` into its output
directory. `CollectExternalCallTargets` emits `jal`/`j` targets that land outside all of this unit's
recompiled, non-stub, non-skipped functions, dropping only targets inside this unit's own data/bss
sections — a mis-decoded `jal` into non-code bytes. A caller cannot know a callee unit's section
layout, so emission is a permissive candidate collector and the ingesting unit decides.
`findContainingFunction` accepts a candidate only at a decoded instruction boundary inside a
recompiled, non-stub, non-skipped, non-entry function it does not start. A new
`[general] external_call_target_manifests` names the sibling manifests to ingest.

**Two-phase build.** `recompile(bool emitManifestOnly = false)` — the default preserves existing
call sites — plus a `--emit-manifest-only` flag. Emission runs at the top of `recompile()` in both
modes and reads only this unit's own decode, so analysis order does not matter. Run every unit's
analysis phase, then every unit's generate phase. A configured manifest that cannot be opened at
generate time is now a hard error reaching a non-zero exit, replacing a warn-and-continue.

**Thread entries.** `DiscoverDataEmbeddedThreadEntries` locates `CreateThread` sites: an inline
`syscall`, or a `jal` to a wrapper whose opening instructions set `$v1 = 0x20`. It recovers the
static `ThreadParam*` in `$a0` through a bounded `lui`/`addiu`/`ori` walk that includes the delay
slot, then reads the entry pointer at `param+4`.

## Basis

- `jal`/`j` encode an absolute 26-bit target, so a cross-unit call target is recoverable from the
  caller's decode alone, with no relocation or symbol.
- `CreateThread` is syscall `0x20` in `$v1`, its `ThreadParam*` argument arrives in `$a0`, and the
  struct's entry pointer is the word at `+4`.
- `mayWriteGprs` derives writable registers from opcode (and `SPECIAL` function) rather than
  decoder-populated booleans, so hand-built and decoded instructions behave alike. Stores read `rt`
  and write no GPR; `COP0`/`COP1`/`COP2`/`MMI` invalidate both `rt` and `rd` rather than sub-decode
  `fmt`. Over-invalidating costs a missed entry, never a stale value.
- The constant walk and the `$v1` back-scan stop at the call site's basic-block boundary. A basic
  block has no interior control transfer, so no value survives a branch or an unrelated call.

## Testing

```
cmake --build build --target ps2x_tests
./build/ps2xTest/ps2x_tests
```

Read the runner's `Failed:` line rather than a pasted count; each test below runs by name.

<details>
<summary>Test roster: two-ELF cross-unit regression, two-phase build and hard-fail guards, target-selection cases, thread-entry def-model and basic-block pins, wrapper-scan reset pair, manifest parsing</summary>

- `two-ELF: A emits T, B ingests T` — drives two independent `PS2Recompiler` instances over two ELF
  fixtures. A's `jal` targets `T`, which exists only inside B and lies outside every section of A.
  A's analysis phase emits `T`; A's manifest is asserted byte-identical whether A's analysis runs
  before or after B's; B then ingests A's actual emitted file and registers `T` under its owning
  function. A negative arm asserts `T` is not registered when B configures no manifest.
- `full pipeline: manifest-ingested packed jal-only entry registers into its owning unit` — the
  single-ELF pipeline with and without a manifest; the without-manifest arm reproduces the
  unregistered target.
- `missing configured manifest hard-fails at generate time` / `existing (even empty) configured
  manifest does not hard-fail`.
- `analysis phase never hard-fails on a missing sibling manifest`, and `two-phase clean build of
  mutually-calling units`, which drives two ELFs calling into each other's mid-body targets through
  a full analysis-then-generate cycle for both.
- `collect external call targets: …` — a foreign/overlay target outside all sections is collected; a
  target inside the caller's own data section is excluded; a `jal` into a local recompiled function
  is excluded while one outside all local functions is included; targets inside skipped or stub
  local functions are still emitted; conditional branches are not collected; duplicates collapse and
  results sort.
- `data-embedded thread entries: …` — `store between materialization and call does not clobber
  $a0`, `strong load-clobber pin`, `unrelated jal between materialization and CreateThread is not
  resolved`, `branch between materialization and inline CreateThread is a block boundary`, `mtc1
  reading $a0 is a documented safe-direction false negative`, `wrapper scan resets on a later $v1
  clobber` with its positive half-guard, `wrapper with wrong syscall number is not registered`,
  `zero entry pointer is filtered out`, `multiple call sites to the same struct dedupe`, `addiu LO
  with sign bit set is sign-extended, not OR'd`, plus the wrapper and inline resolution cases and
  the `jal` decoder-populated `rt` case.
- `full pipeline: data-embedded thread entry (CreateThread ThreadParam) registers into its owning
  unit`.
- `external call target manifest parsing sorts, dedupes, and ignores comments/blanks` and `…handles
  empty input`.

</details>

## Risk and not in scope

- Every recompile now writes `external_call_targets.txt` into its output directory, including
  single-unit builds that ingest nothing — one extra file for integrators packaging that output.
- Ingestion is opt-in through config; the thread-entry scan is not, and runs on every recompile. Its
  failures warn rather than abort the pass.
- The hard failure on an unreadable manifest only affects configs setting
  `external_call_target_manifests`.
- Not covered by the thread-entry scan: `StartThread` (`0x22`), runtime-computed `ThreadParam`
  pointers, handler-registration variants. An `mtc1 $a0` before the call over-invalidates and loses
  the entry, asserted as a known false negative.
- A single `addiu $v1,$zero,0x20` inside the back-scan window can be attributed to two syscalls; the
  extra candidate is de-duplicated downstream.
- Unrelated to the feature: `<cstdint>` is added ahead of the elfio include, fixing the build on
  newer GCC.


ps2xRecomp/include/ps2recomp/ps2_recompiler.h
```diff
@@ -46,6 +48,27 @@ namespace ps2recomp
...
+        // values read from data memory, sorted and de-duplicated.
+        static std::vector<uint32_t> DiscoverDataEmbeddedThreadEntries(
+            const std::unordered_map<uint32_t, std::vector<Instruction>> &decodedFunctions,
+            const std::function<bool(uint32_t)> &isValidAddress,
+            const std::function<uint32_t(uint32_t)> &readWord);
+
+        // Collects jal/j targets that fall in executable sections but outside every
+        // recompiled local function range - candidate cross-unit call targets to
+        // publish in the external call-target manifest. Sorted and de-duplicated.
+        static std::vector<uint32_t> CollectExternalCallTargets(
```
ran-j, line 77: This method currently requires the target to be inside an executable section of the caller's ELF. This appears to exclude the main use case described by the PR: a direct jal from a main ELF to a separately compiled overlay whose target range is not represented by any executable section in the main ELF.

The current unit test explicitly expects targets outside the current ELF's executable sections to be discarded, while the end-to-end test starts from a manually pre-created manifest and therefore never exercises cross-ELF manifest emission.

Could we add a real two-ELF regression test where ELF A contains a jal to code that exists only in ELF B, then verify that A emits the target and B ingests it? Unless both ELFs happen to have overlapping executable section ranges, the current filter seems to drop exactly the cross-unit target this feature is intended to preserve.

ps2xRecomp/src/lib/ps2_recompiler.cpp
```diff
@@ -982,6 +984,7 @@ namespace ps2recomp
 #endif
             }
 
+            loadExternalCallTargetManifests();
             discoverAdditionalEntryPoints();
```
ran-j, line 1004: There seems to be a build-order problem here. Each invocation loads sibling manifests before discoverAdditionalEntryPoints(), but only emits its own manifest at the end of that pass.

For mutually calling units, such as a main ELF and an overlay that call into each other, a clean build creates a dependency cycle: each output requires the other unit's manifest before that manifest has been produced. The first invocation only logs a warning and continues, leaving its generated registration table incomplete until it is run again.

The end-to-end test pre-creates the manifest, so it does not exercise this clean-build scenario. Could manifest generation be split into a separate analysis phase, or could the project provide an explicit deterministic two-pass orchestration? At minimum, silently continuing when a configured manifest is missing seems unsafe because it produces valid-looking but incomplete output.

ps2xRecomp/src/lib/ps2_recompiler.cpp
```diff
@@ -2120,4 +2302,297 @@ namespace ps2recomp
...
+        }
+
+        // Simplified "who writes this register" model shared by the $v1 clobber check
+        // and the $a0 constant-propagation walk below: for I-type opcodes the written
+        // register is rt, for OPCODE_SPECIAL it is rd. $zero is never considered written.
+        // J-type instructions (j/jal) are excluded: the decoder unconditionally populates
+        // rs/rt/rd from the raw instruction bits (see R5900Decoder::decodeInstruction),
+        // but for a 26-bit jump target those bit positions are part of the target, not a
+        // register field, so treating them as a register write would be spurious.
+        uint32_t writtenRegisterOrZero(const Instruction &inst)
```
ran-j: this method treats `rt` as the destination of every non-SPECIAL, non-J-type instruction. That is not valid for stores, branches, and several coprocessor instructions. For example,` sw $a0`, ... reads $a0 but would currently invalidate the resolved $a0, causing a false negative.

There is also a false-positive direction: previous jal instructions are treated as writing no tracked register, so a resolved $a0 can survive across an arbitrary function call even though $a0 is caller-saved and may have been clobbered by the callee. The linear propagation walk also crosses control-flow instructions without checking whether the materialization dominates the CreateThread call.

Could this use an opcode-aware register-def helper and stop or invalidate state at earlier calls and control-flow boundaries? Restricting the analysis to the call site's basic block plus the delay slot would be more conservative.

Please add regression tests for at least:

```
lui $a0, ...
sw $a0, 0($sp)
jal CreateThread
addiu $a0, $a0, ...
```

and for an unrelated jal between the $a0 materialization and the CreateThread call.

ps2xRecomp/src/lib/ps2_recompiler.cpp
```diff
@@ -2120,4 +2302,297 @@ namespace ps2recomp
...
+    {
+        std::vector<uint32_t> results;
+
+        // Step 1: identify CreateThread (syscall 0x20) wrapper function entry points -
+        // functions whose first few instructions set $v1 = 0x20 and execute a syscall.
+        std::unordered_set<uint32_t> wrapperStarts;
+        for (const auto &[functionStart, instructions] : decodedFunctions)
+        {
+            const size_t window = std::min(instructions.size(), kThreadWrapperScanWindow);
+            bool sawAddiuV1Syscall = false;
```
ran-j, line 2582: sawAddiuV1Syscall becomes true, the wrapper scan accepts any later syscall in the scan window, even if $v1 was overwritten in between.

For example, this is currently classified as a CreateThread wrapper even though the executed syscall number is 0x21:

```
addiu $v1, $zero, 0x20
addiu $v1, $zero, 0x21
syscall

```
That can cause ordinary calls to the function to be interpreted as CreateThread calls and arbitrary memory to be read as a ThreadParam.

Could the wrapper scan apply the same $v1 clobber-aware logic used by the inline-syscall scan and include a negative regression test for this sequence?

ran-j (changes requested):

smmathews:
All four points were correct, thanks especially for catching that the cross-ELF emit path was never actually exercised. Reworked (follow-up commits; body updated):

- Emission: CollectExternalCallTargets is now a permissive candidate collector (it excludes only intra-unit targets and targets inside the caller's own data/bss sections); the ingest side's findContainingFunction remains the authoritative filter. The test suite now has the real two-ELF regression you asked for: ELF A jal's a target that exists only in ELF B, A's emitted manifest is asserted to contain it, and B ingests A's actual emitted file.
- Build order: manifest emission is split into an analysis phase (--emit-manifest-only), so mutually-calling units do analysis-all-then-generate-all with no cycle, emission depends only on the unit's own decode, so this is deterministic with no fixpoint. A configured manifest that's missing at generate time is now a hard failure, not a warning.
- Register-def model: replaced the rt-is-always-destination assumption with an opcode-keyed conservative helper (stores/branches don't write; loads do; COP conservatively over-invalidates in the safe direction), and the constant-propagation walk is bounded to the call site's basic block plus the delay slot, which also kills value survival across unrelated calls. Your exact sw $a0 and unrelated-jal sequences are regression tests.
- Wrapper scan: sawAddiuV1Syscall now resets on any later $v1 clobber, same logic as the inline scan; your 0x20/0x21 sequence is the negative test.

PR #151, opened 2026-07-07
feat(runtime): deferred INTC pending-raise latch with per-tick drain

smmathews:
## Problem

`submitDmaSend` completes the VIF1 transfer synchronously through `processPendingTransfers()` but never raises INTC cause 5. A guest using the sce-libdma kick-and-wait shape — `sceDmaSend`, then `AddIntcHandler` / `EnableIntc`, then `WaitSema` — is never released and blocks forever. Observed on Dragon Quest VIII's per-frame VIF1 display-list submit path.

The handler is registered *after* the kick returns, so dispatching at the raise site would deliver to nothing.

## Fix

- **`ps2_vif1_interpreter.cpp`, `ps2_memory.{h,cpp}`** — `processVIF1Data` already decodes the VIFcode I bit (`cmd & 0x80000000`); that `if (irq)` site now also records an edge on `PS2Memory`, read back through a new `consumeVif1InterruptEdges()` mirroring `consumeCompletedDmacCauses()`.
- **`Interrupt.cpp` / `Interrupt.h`** — a level-triggered pending latch. `raisePendingIntc(cause)` sets one bit per cause from any thread and stamps a monotonic timestamp on every raise; causes ≥ 32 and vblank causes 2 and 3 are ignored (vblank stays the worker's own periodic dispatch). `drainPendingIntc(rdram, runtime)` runs once per worker tick and clears a bit only after a registered, enabled handler was dispatched, or once the raise outlives the expiry window (~2 s). `dispatchIntcHandlersForCause` becomes `dispatchAndCountIntcHandlersForCause`, returning the dispatch count so the drain can tell delivery from a miss.
- **`Support.h`** — `submitDmaSend` raises cause 5 only on a non-zero consumed edge count, leaving delivery to the next drain tick.
- **`ps2_runtime.cpp`** — `drainCompletedDmacHandlers` consumes edges too, so I-bit VIFcodes parsed outside `submitDmaSend` (GS / Font / GIF-chain paths) are delivered in the same frame, not leaked into a later kick.

## Basis

- The VIFcode I bit is the VIF1 interrupt assert, decoded already where the interpreter sets `STAT.INT`. Raising on every kick would invoke a persistent cause-5 handler for ordinary packets, signalling semaphores or advancing display-list state early.
- Expiry uses `steady_clock`, not a drain-tick count, so a cause raised before anything starts the interrupt worker cannot outlive the window. `AddIntcHandler` starts that worker, and the target sequence raises before it.
- Freshness is measured from the most recent raise, so a re-asserted cause is not dropped on an older assertion's age.

## Testing

`ps2xTest/src/ps2_runtime_interrupt_tests.cpp`, worker stopped so drain ticks are deterministic:

- An I-bit VIFcode kicked through `sceDmaSend` sets the pending bit before any handler exists; a handler registered afterwards runs once with cause 5 in `$a0`, and the bit clears.
- The same kick without an I bit sets no bit, and a cause-5 handler registered afterwards never runs.
- A raise made while a handler is already registered is delivered on the first drain; a raise made with no handler yet survives that drain and is delivered on the next one after registration.
- Under an injectable clock: a pending cause survives just under the window, a re-raise refreshes its timestamp, and it drops only once the window elapses from that latest raise.
- Under the same clock, a raise predating any worker or drain is dropped, not delivered, on the first drain it sees.
- `raisePendingIntc` on causes 2, 3 and 32 leaves the latch at zero.

## Risk and not in scope

- Delivery and expiry both need a `drainPendingIntc` call. A cause raised while nothing drives a drain stays pending until one runs; no timer thread expires it.
- `Interrupt.h` newly exposes `drainPendingIntc`, `resetInterruptHandlerState`, `setPendingIntcClockForTest` and `g_pending_intc_causes` for tests.
- Only cause 5 is raised today (`git grep -n 'raisePendingIntc(' -- ps2xRuntime/src`); the latch itself is generic.
- PR #190 also changes `ps2_memory.cpp` and `ps2_vif1_interpreter.cpp`: different functions in the former, both inside `processVIF1Data` in the latter.
- Also carried: `#include <cstdint>` ahead of `elfio.hpp` in `elf_parser.h`, to build the recompiler against newer libstdc++. Unrelated to INTC; happy to split out.

<details>
<summary>Build and test commands, the same-cause raise collapse, why <code>resetInterruptHandlerState</code> has to live in <code>Interrupt.cpp</code>, and the <code>std::countr_zero</code> choice</summary>

Built and run clean, matching the Linux CI job:

```
cmake --build build --config Release
./build/ps2xTest/ps2x_tests
```

The full suite passes.

Two raises of one cause landing either side of a drain's clear collapse into a single delivery. Under a level-triggered latch the bit means "at least one pending", not a count; the comment at the `fetch_and` site records that.

`resetInterruptHandlerState()` is declared in `Interrupt.h` and defined in `Interrupt.cpp`. It restores the INTC/DMAC handler tables, enable masks, id and order counters, the pending latch and the real clock. It has to live in that translation unit: `g_intcHandlers` and `g_dmacHandlers` are `static` at namespace scope in `Syscalls/Helpers/Runtime.h`, so every including TU holds its own copy, and the raise-timestamp array is not exposed in the header at all. Tests call it from `cleanupRuntime` for INTC isolation; function-table isolation is separate, from `MiniTest::BeforeEach(reset_ps2_test_function_table)` in `ps2xTest/src/main.cpp`.

`drainPendingIntc` uses `std::countr_zero` (C++20) rather than `__builtin_ctz` so the MSVC job builds.

</details>


ps2xRuntime/src/lib/Kernel/Syscalls/Interrupt.cpp
```diff
@@ -216,6 +225,98 @@ namespace ps2_syscalls
...
+                const uint32_t logIndex = s_raiseLogCount.fetch_add(1u, std::memory_order_relaxed);
+                if (logIndex < 16u || (logIndex % 256u) == 0u)
+                {
+                    RUNTIME_LOG("[INTC:raise] cause=" << cause);
+                }
+            });
+        }
+    }
+
+    void drainPendingIntc(uint8_t *rdram, PS2Runtime *runtime)
```
ran-j, line 275: This unconditional raise can change guest synchronization semantics, not just interrupt timing. A game may keep a cause-5 handler registered while submitting many ordinary VIF1 packets and request an interrupt only on selected VIFcodes using the I bit. With the current implementation, every packet invokes that handler and may signal semaphores or advance display-list state prematurely.

The VIF1 interpreter already has the exact per-command edge available:

const bool irq = (cmd & 0x80000000u) != 0u;

I agree that reading the sticky STAT.INT bit later from submitDmaSend() would be unreliable, but we do not need to use the sticky register as the event source. Could processVIF1Data() record an interrupt edge when it decodes an I bit, and could PS2Memory expose those events through a consume API similar to consumeCompletedDmacCauses()?

Please add integration coverage for both cases:

A VIF1 transfer containing a VIFcode with I=1 raises cause 5.
A normal VIF1 transfer with no I bit does not raise cause 5.

The current over-approximation is unsafe whenever a persistent cause-5 handler exists.

ps2xRuntime/src/lib/Kernel/Syscalls/Interrupt.h
```diff
@@ -24,9 +26,16 @@ namespace ps2_syscalls
         extern uint32_t g_enabled_dmac_mask;
         extern uint64_t g_vsync_tick_counter;
         extern VSyncFlagRegistration g_vsync_registration;
+        extern std::atomic<uint32_t> g_pending_intc_causes;
+        constexpr uint32_t kPendingIntcMaxAgeTicks = 120u;
     }
 
     void dispatchDmacHandlersForCause(uint8_t *rdram, PS2Runtime *runtime, uint32_t cause);
+    void raisePendingIntc(uint32_t cause);
```
ran-j, line 36: raisePendingIntc() only sets the pending bit. The age counter advances exclusively from drainPendingIntc(), and the production drain is called by the interrupt worker.

This means the advertised approximately two-second expiration does not apply before anything has started the worker. A VIF1 kick can set cause 5, no handler or VSync API may be used for an arbitrary amount of time, and a handler registered much later can receive that stale interrupt because its age remained zero the entire time.

This is particularly relevant because the targeted sequence raises before AddIntcHandler(), and AddIntcHandler() is what starts the worker.

Could expiration use a monotonic timestamp associated with the raise, or could the implementation otherwise guarantee that pending causes are drained even when no handler/VSync worker existed at raise time? Please add a regression that covers a raise occurring before worker startup and verifies that it cannot survive beyond the configured expiration window.

ps2xRuntime/src/lib/Kernel/Syscalls/Interrupt.cpp
```diff
@@ -216,6 +225,98 @@ namespace ps2_syscalls
...
+
+    void raisePendingIntc(uint32_t cause)
+    {
+        if (cause >= 32u || cause == kIntcVblankStart || cause == kIntcVblankEnd)
+        {
+            return;
+        }
+
+        const uint32_t bit = 1u << cause;
+        const uint32_t prev = g_pending_intc_causes.fetch_or(bit, std::memory_order_acq_rel);
```
ran-j, line 261: The age is reset only when fetch_or() observes a 0 -> 1 transition. If a cause has been pending for almost the full timeout and the same cause is raised again, the second raise does not refresh the age because the bit is already set. The next drain can therefore drop an interrupt that was asserted only one tick ago.

Collapsing multiple raises into one pending level is reasonable, but the artificial stale-event timeout should not classify a recently reasserted level using the age of an older assertion.

Please add a regression that:

Raises a cause.
Advances it to the age-out boundary.
Raises the same cause again while the bit is still set.
Verifies that the next drain does not discard the newly reasserted event.

Once cause 5 is raised only for real VIF I events, resetting the age on each raise should no longer conflict with expiration.

ps2xTest/src/ps2_runtime_interrupt_tests.cpp
```diff
@@ -887,5 +905,162 @@ void register_ps2_runtime_interrupt_tests()
 
             cleanupRuntime(env);
         });
+
+        tc.Run("raisePendingIntc delivers to a registered handler on the next drain tick", [](TestCase &t)
```
ran-j, line 1026: All new tests drive raisePendingIntc() directly, so they validate the latch but not the behavior introduced in submitDmaSend().

Please add an integration test that submits an actual VIF1 packet, lets processPendingTransfers() parse it, registers/enables the handler after the kick, and then verifies deferred delivery. A corresponding no-I packet test should verify that no cause-5 interrupt is produced.

Without those tests, the suite cannot detect the unconditional producer bug or prove that the DQ8-style DMA-to-late-handler path is what actually unblocks.

ran-j (changes requested):

smmathews:
Thanks for the thorough review, all four points were right, and I've reworked the PR accordingly (pushed as follow-up commits so your line comments stay anchored; the PR body is updated to match):

- Edge-based raise: processVIF1Data now records an interrupt edge at the exact if (irq) site when it decodes an I-bit VIFcode, exposed via PS2Memory::consumeVif1InterruptEdges() mirroring consumeCompletedDmacCauses(). submitDmaSend raises cause 5 only on a consumed edge — no more unconditional raise. A second consume site in the runtime's DMAC-drain path delivers edges from VIF1 packets parsed outside submitDmaSend within the frame.
- Expiration: switched from drain-tick aging to a monotonic wall-clock timestamp per cause (~2 s window), so a raise before anything starts the worker can't survive indefinitely. The clock is injectable for deterministic tests.
- Age refresh: the timestamp now refreshes on every raise, not just the 0→1 transition, so a recently reasserted cause can't be dropped on an old assertion's age.
- Tests: both integration tests you asked for are in: a real VIF1 packet with I=1 through sceDmaSend→processPendingTransfers with the handler registered after the kick, and a no-I negative, plus regressions for re-raise-at-the-age-out-boundary and raise-before-worker-startup expiry.

PR #154, opened 2026-07-07
feat(hle): parameterize SoundDriver RPC service SID + subcommands per game

smmathews:
## Problem

The `TSNDDRV` module answers a fixed pair of service ids and a fixed set of subcommand
numbers, held as file-local `constexpr` values in `tsnddrv.cpp` (`kCommandSid`,
`kStateSid`, `kSubmitFunction`, `kGetStatusAddressFunction`,
`kGetAddressTableFunction`). Its other title-specific numbers already arrive through
`TsnddrvBindings` from the game's entry in `builtin_profiles.cpp`: arena geometry,
checksum candidates, busy flag address, completion rules. A title whose driver registers
a different service id, or numbers its subcommands differently, cannot reuse the module
without editing it.

## Fix

- `TsnddrvBindings` gains `commandSid`, `stateSid`, `submitFunction`,
  `getStatusFunction`, `getAddrTableFunction`. Defaults are the deleted constants' values,
  so the shipped profile behaves as before.
- `handleRpc` dispatches against the bindings, and `sids()` is built at construction, so a
  driver that muxes command and state onto one service id registers it once, not twice.
- Optional subcommands, disabled at `0`: `streamOpenFunction` writes
  `streamReadyValue` to `streamStateAddress`; `channelConfigFunction` sets
  `channelAllocFlagTableAddress[channel] = 1`, taking `channel` from send word 0 when
  below 16; `stopFunction` writes `1` to `stopCompletionFlagAddress`. Each zeroes the
  receive word and signals nowait completion.
- A subcommand the module does not recognize, on a service id it owns, writes
  `benignStatusValue` (default `0xffffff9b`) to the receive word and stays **unhandled**,
  so a guest-registered server function at that service id still runs.
- `createTsnddrvService` throws `std::invalid_argument` when `getStatusFunction` equals
  `getAddrTableFunction`, or when an enabled optional subcommand aliases submit,
  getStatus or getAddrTable — either would make dispatch ambiguous.
- The `recvx-us` profile (`slus_201.84`) states its service ids and subcommand numbers
  explicitly rather than inheriting them.

## Basis

The numbers are the deleted constants, carried over unchanged: the removed `constexpr`
block and the profile's new fields are both in this diff. `TsnddrvBindings` is already
how the module takes its per-title addresses.

## Testing

New cases in `ps2_iop_tests.cpp`:

- A synthetic profile muxing command and state onto `0x80007701` with subcommands
  `0x30`/`0x31`/`0x32`: `sids()` reports that id once, the custom getStatus returns a
  nonzero status address, and a request on `1` is refused.
- Optional subcommands at `0x40`/`0x41`/`0x42`: byte-level checks that `streamReadyValue`
  reaches `streamStateAddress`, that channel 5's slot in `channelAllocFlagTableAddress`
  becomes `1`, and that `stopCompletionFlagAddress` becomes `1`. Subcommand `0x77`
  returns unhandled with `benignStatusValue` in the receive word.
- Both rejected binding shapes throw.

The sound-status cases in `ps2_sif_rpc_tests.cpp` spell `1`, `0x12` and `0x13` at the
point of use instead of that file's `IOP_SID_SNDDRV_*` / `IOP_RPC_SNDDRV_*` aliases,
which no longer track the module. `ps2xTest/CMakeLists.txt` adds `ps2xIOP/src` to the
test include path.

## Risk and not in scope

- Behavior change for `slus_201.84`: an RPC on service id `0` or `1` carrying an
  unrecognized subcommand, with a receive buffer of at least four bytes, now gets
  `0xffffff9b` written to its receive word before falling through. `SifCallRpc`'s
  finalization in `Kernel/Syscalls/RPC.cpp` then overwrites that word — copy from send,
  or zero — because the call stays unhandled. It survives only when a guest server
  function at that id is dispatched and returns `0` with the server buffer at `0`,
  leaving the result pointer equal to the receive buffer.
- The optional subcommands match on subcommand number alone, not on which of the two
  service ids the call arrived on. Distinct numbers disambiguate in real drivers, and the
  muxed single-id case depends on it.
- `submitFunction` colliding with `getStatusFunction` or `getAddrTableFunction` is not
  rejected; they sit on different service ids in the shipped profile.
- No new title profile is added. The SPU2 transfer-handle pairs and stream
  descriptor a second title would need have no subcommand slot, no trigger and no tests.
- `ps2xRecomp/include/ps2recomp/elf_parser.h` gains `#include <cstdint>`, unrelated to
  sound. That fix landed separately in #199, so the line drops out of this branch on the
  next rebase.

<details>
<summary>TsnddrvBindings SID and subcommand fields — name, default, effect — and the values the recvx-us (slus_201.84) profile sets</summary>

| Field | Default | Effect |
|---|---|---|
| `commandSid` | `0x00000000` | service id carrying submit |
| `stateSid` | `0x00000001` | service id carrying the status / addr-table queries |
| `submitFunction` | `0x00000000` | submit command buffer |
| `getStatusFunction` | `0x00000012` | return `statusAddress` |
| `getAddrTableFunction` | `0x00000013` | return `addressTableAddress` |
| `streamOpenFunction` | `0` (off) | write `streamReadyValue` to `streamStateAddress` |
| `channelConfigFunction` | `0` (off) | set `channelAllocFlagTableAddress[channel] = 1` |
| `stopFunction` | `0` (off) | write `1` to `stopCompletionFlagAddress` |
| `benignStatusValue` | `0xffffff9b` | receive word for an unrecognized subcommand |
| `streamStateAddress` | `0` | target of the streamOpen write |
| `streamReadyValue` | `0` | value streamOpen writes |
| `channelAllocFlagTableAddress` | `0` | base of the per-channel allocation flags |
| `stopCompletionFlagAddress` | `0` | target of the stop write |

`0` disables the three optional subcommands. It does not disable a service id: `0` is a
real service id here, and `slus_201.84` uses it.

`recvx-us` (`slus_201.84`) sets:

```
.commandSid           = 0x00000000
.stateSid             = 0x00000001
.submitFunction       = 0x00000000
.getStatusFunction    = 0x00000012
.getAddrTableFunction = 0x00000013
```

It leaves the optional subcommands and their addresses at their defaults, so its
dispatch is unchanged apart from the benign-status write noted above.

</details>


ran-j:
Hey,  sorry but I broke your PR

smmathews:
> Hey, sorry but I broke your PR

no worries, I'm on it

PR #178, opened 2026-07-21
feat(mpeg): host-fed CD-stream ES injection for the guest-driven decoder

smmathews:
## Problem

The guest-driven MPEG decoder accepts elementary stream only through the guest's own
demux syscalls (`sceMpegDemuxPss`, `sceMpegDemuxPssRing`), which read from guest RAM at a
guest pointer. Some runtimes HLE the movie stream's IOP-side delivery, where an IOP
streamer fills an EE ring the runtime does not model. Such a runtime holds a correct,
in-order program stream in host memory while the guest's `sceMpegGetPicture` consumer is
live, but has no supported way to hand those bytes to the decoder. It must instead
reimplement the guest ring protocol or carry a second decoder. The same gap applies when a runtime applies an
on-disc stream transform host-side before decode.

## Fix

Add `ps2_stubs::feedMpegCdStreamBytes(const uint8_t *, size_t)`, the data entry point to
go with the existing `notifyMpegCdStreamStart` / `notifyMpegCdStreamEof` control calls.

- Routes host bytes into every playback state on the current CD-stream generation and
  parses them with `appendPssBytes`, the same PSS path `sceMpegDemuxPss` feeds.
- Takes no `mpegAddr`: a host streamer does not know the guest handle, so routing uses
  the existing generation bookkeeping.
- Order-insensitive with respect to `sceMpegCreate`. Bytes fed while no decoder exists on
  the current generation are staged in host memory and replayed into the next decoder
  created on that generation, before its first `getPicture`.
- The stage is bounded by `kMpegHostFeedStageCapBytes`. On overflow the whole stage is
  dropped rather than truncated, because a truncated stage would hand a later decoder a
  gapped stream; staging is then abandoned for the rest of that generation and the next
  decoder resyncs on live feed. Staged bytes are also dropped at stream stop and restart.
- Returns `size` whenever a CD stream is active, routed or staged. Returns 0 when no
  stream is active, and for a null or empty input. It is not a decoded-byte count and
  does not scale with how many decoders are open.

Supporting changes: staging fields on the MPEG stub state, with the staged buffer and its
overflow flag cleared by `resetMpegStubState` and by both `notifyMpegCdStream*` calls.
Also a replay block in `sceMpegCreate`, and a defaulted `trackGuestAddrs` parameter on
the internal `appendPssBytes` that leaves the guest demux and callback paths unchanged.

## Basis

Symmetric with, and reusing, the merged `notifyMpegCdStream*` contract already in this
tree. Demux goes through the existing PSS/PES path rather than a second parser. No game
knowledge enters the runtime: no file layout, no stream transform, no addresses.

## Testing

```
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS=-msse4.1 -DCMAKE_CXX_FLAGS=-msse4.1 && cmake --build build && ./build/ps2xTest/ps2x_tests
```

Suite passes. The decode assertions need an FFmpeg-enabled build
(`PS2X_ENABLE_FFMPEG`, on by default off Android).

<details>
<summary>Added tests: routing round-trip, two-decoder fan-out, feed-before-start no-op, staged replay, split packet, CD-stream restart, stage-cap overflow and per-generation recovery, stage discard at EOF — plus mutation-coverage notes</summary>

All in `ps2xTest/src/ps2_runtime_expansion_tests.cpp`:

- `host-fed CD stream decodes without a guest demux call` — a create / start / feed / eof
  / getPicture round-trip with no `sceMpegDemuxPss*` call in the path.
- `host feed fans out to every open decoder and reports the input range once` — the
  return value is the input range, not once per decoder, and both open decoders on the
  generation receive the bytes.
- `feedMpegCdStreamBytes before a CD stream start is a no-op` — handle exists,
  generation 0, returns 0.
- `host feed staged before a decoder opens reaches the later-created decoder` — the feed
  reports as accepted and the decoder created afterwards decodes it.
- `host feed split across a packet boundary decodes`.
- `host feed decodes across a CD-stream restart`.
- `host feed staging is bounded and recovers after overflow` — past the cap the staged
  prefix is discarded (width stays the seeded default), and a fresh generation stages and
  decodes normally again.
- `host feed staged before EOF is discarded, not resurrected after EOF` — a decoder
  created after EOF does not replay a dead generation's prefix.

Mutation coverage: each property these tests pin has a source-level mutation that fails
it. Some properties share pre-existing partial-buffer machinery and fail as a group
rather than singly, and one defensive restart guard has no public-API mutation that
isolates it. Full matrix on request.

</details>

## Risk and not in scope

- No in-tree caller. Nothing in `ps2xRuntime` invokes the new function, and it returns
  early when no CD stream is active, so existing behaviour is unaffected. It exists for
  embedders of the runtime library that produce the elementary stream host-side.
- Stream callbacks are not dispatched on this path. A host feed has no calling guest
  thread, and host-fed bytes record no guest addresses, so no callback event is queued
  (marked `// TODO(host-feed callbacks)`). If a title needs them, the follow-up is a
  per-`MpegPlaybackState` event queue drained by the next guest MPEG syscall. Flagged
  for your call.
- One decoder per generation is the supported shape. A second decoder opening after
  bytes have already routed to the first gets no replay and resyncs at the next sequence
  header, as with guest demux.
- `kMpegHostFeedStageCapBytes` is an internal tuning value, not part of the API contract.
- The suspected bug reported on this PR is still unconfirmed; nothing above claims it is
  resolved.

### DIFF
--- a/ps2xRuntime/src/lib/Kernel/Stubs/MPEG.cpp
+++ b/ps2xRuntime/src/lib/Kernel/Stubs/MPEG.cpp
@@ -502,6 +502,9 @@ namespace ps2_stubs
             uint32_t nextCallbackHandle = 1u;
             uint64_t cdStreamGeneration = 0u;
             bool currentCdStreamEofSeen = false;
+            std::vector<uint8_t> cdStreamStagedBytes;       // host-fed ES held before a decoder exists
+            uint64_t cdStreamStageGeneration = 0u;          // generation the stage belongs to
+            bool cdStreamStageOverflowed = false;           // cap hit -> stage abandoned this generation
             uint32_t feedEsTraceCount = 0u;
             uint32_t demuxPssTraceCount = 0u;
             uint32_t demuxRingTraceCount = 0u;
@@ -1154,7 +1157,8 @@ namespace ps2_stubs
                             const uint8_t *data,
                             size_t size,
                             uint32_t guestAddr,
-                            std::vector<MpegStreamCallbackEvent> &callbackEvents)
+                            std::vector<MpegStreamCallbackEvent> &callbackEvents,
+                            bool trackGuestAddrs = true)
         {
             if (!data || size == 0)
             {
@@ -1184,10 +1188,13 @@ namespace ps2_stubs
             playback.sawInput = true;
 
             playback.pssBuffer.insert(playback.pssBuffer.end(), data, data + size);
-            playback.pssGuestAddrs.reserve(playback.pssGuestAddrs.size() + size);
-            for (size_t i = 0; i < size; ++i)
+            if (trackGuestAddrs)
             {
-                playback.pssGuestAddrs.push_back(guestAddr + static_cast<uint32_t>(i));
+                playback.pssGuestAddrs.reserve(playback.pssGuestAddrs.size() + size);
+                for (size_t i = 0; i < size; ++i)
+                {
+                    playback.pssGuestAddrs.push_back(guestAddr + static_cast<uint32_t>(i));
+                }
             }
             processPssBuffer(mpegAddr, playback, callbackEvents);
             if (!playback.decodedFrames.empty())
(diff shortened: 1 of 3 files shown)


ran-j:
I think I found a but on this PR but still checking

smmathews:
While auditing our side I did find one concrete bug, in case it's the same thing you hit: feedMpegCdStreamBytes only routed to decoders that already existed on the current CD-stream generation, so bytes fed after notifyMpegCdStreamStart but before the guest's first sceMpegCreate were silently dropped (return 0, no staging), and a decoder opened later never recovered that prefix. The return value also couldn't distinguish "no active stream" from "active but no decoder yet."

I've fixed that in this PR (just pushed, rebased on current main): host-fed bytes now stage at the generation level (bounded, discarded on overflow/stop/restart rather than truncated) and replay to the next decoder that opens, and the return contract is now "size whenever a CD stream is active, routed or staged; 0 only when inactive", with regression tests for feed-before-open, the staging cap, and stage discard. If what you found is something else, the details would be very welcome.

PR #181, opened 2026-07-22
Generalize dispatchGuestBranch resume handling + several boot/IOP hang fixes

LuisFellp:
## Summary

A set of runtime fixes found while bringing up a new game (Taiko no Tatsujin, SLPS-20414) far enough to reach real gameplay logic. All of these are general runtime bugs, not game-specific hacks (the two game-compat entries are clearly scoped through the existing `game_overrides.cpp` mechanism).

- **`dispatchGuestBranch` resume-loop generalization** (`ps2_runtime.cpp`): a callee can legitimately leave `ctx->pc` at an internal resume label instead of the expected fallthrough, because Ghidra sometimes splits one continuous piece of guest code into several recompiled functions (tail calls, jump tables, branch targets not recognized as their own entry point). The old code only re-drove the *original* callee; if that callee tail-called into a different function needing resuming, dispatch gave up too early. This generalizes it to keep re-driving whatever function `ctx->pc` currently belongs to, with a bounded guard against runaway spins. This mirrors the intent of the unmerged `#115`/`#117` PRs, generalized further.
- **COP0 `Status.IE` never set at boot** (`ps2_runtime.cpp`, `Thread.cpp`, `Interrupt.cpp`): real PS2 BIOS/IPL sets this before ever handing control to the game's entry point; since this recomp skips BIOS/IPL, nothing ever set it, which silently blocked every single `CreateThread`'d thread from ever reaching `StartThread` (common libkernel helper code checks "were interrupts enabled" first). Fixed at all three places a fresh `R5900Context` gets its initial state.
- **`sceSifSetDma` hang on `dst==0` queued `SIF_CMD` packets**, plus two more SIF boot-status/completion gaps (`SIF.cpp`) found while chasing a hang further downstream.
- **Missing `sceSifSetReg`/`sceSifGetReg` dispatch cases** (`Dispatcher.cpp`).
- Low-noise diagnostic logging added to a couple of failure paths (`CreateSema`, `sceCdInit`) that were useful while chasing these down and seemed worth keeping.
- Two `game_overrides.cpp` entries for Taiko no Tatsujin: a general compatibility override, and an experimental override for a texture-decompress halt (documented in its commit as experimental / known to produce placeholder texture data, since the real root cause -- an unreached CD-open path -- is still open).
- A small reusable Ghidra script (`DecompileTargets.java`) that dumps decompiled pseudocode for a given list of addresses to a file.

Deliberately left out: the recompiled function bodies and function-registration table generated for this specific game's ELF -- those aren't general runtime code and don't belong in this template repo.

## Test plan

- [x] `ps2_runtime` (Release) builds clean with these changes on top of current `main`
- [x] Verified live against Taiko no Tatsujin: boot no longer stalls where it used to (games's threads now actually reach `StartThread`, SIF completion no longer hangs on the fixed paths), and a marathon run went from stalling after ~14.6k vblank ticks to running 80k+ ticks with no new dispatch failures
- [ ] No other game re-tested against this branch -- would appreciate a sanity check against whatever games are currently in local regression rotation, since the `dispatchGuestBranch` and `Status.IE` changes are both hit on every boot for every game

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JqRdbWSfD1oX9jDPkrq6iv


LuisFellp:
Closing this for now -- opened it before checking in with the project's community norms around AI-assisted contributions, my mistake. Will revisit through a different process if it still makes sense later.

PR #199, opened 2026-08-01
fix(recomp): include <cstdint> before elfio in elf_parser.h

smmathews:
## Problem

`ps2xRecomp/include/ps2recomp/elf_parser.h` includes `<elfio/elfio.hpp>` without
first including `<cstdint>`. ELFIO — pinned at `Release_3.12` via FetchContent in
`ps2xRecomp/CMakeLists.txt` — uses `uint16_t`, `uint32_t` and `uint64_t` in
`elf_types.hpp` without including `<cstdint>` itself. Newer libstdc++ releases
trimmed the transitive includes that used to supply those typedefs, so a clean
build of `main` fails on a current toolchain:

```
FAILED: ps2xRecomp/CMakeFiles/ps2_recomp_lib.dir/src/lib/elf_parser.cpp.o
build/_deps/elfio-src/elfio/elf_types.hpp:30:20: error: 'uint16_t' does not name a type
   30 | using Elf_Half   = uint16_t;
note: 'uint16_t' is defined in header '<cstdint>'
```

## Fix

Include `<cstdint>` immediately before `<elfio/elfio.hpp>`, with a comment
recording why the order matters. Header-only; no behaviour changes.

## Testing

Reproduced both directions on a clean checkout, GCC 16.1.1:

```
cmake --build build --target ps2_recomp_lib
```

Without the change the target fails with the error above; with it the target
builds and links.

## Risk and not in scope

The include is unconditional and costs nothing on toolchains that still pull the
typedefs in transitively. Bumping the ELFIO pin, or reporting the missing include
upstream to ELFIO, would also fix it — neither is attempted here.

Seven open branches of mine carry this same one-line fix as a passenger, because
none of them could be built without it: #137, #150, #151, #152, #154, #156 and
#157. None of their descriptions mentions it, so a reviewer of any one of them
meets an unexplained header change. If this lands first they all drop it on
rebase; whichever order they merge in, the duplicate is a trivial conflict rather
than a behavioural one.

### DIFF
--- a/ps2xRecomp/include/ps2recomp/elf_parser.h
+++ b/ps2xRecomp/include/ps2recomp/elf_parser.h
@@ -1,6 +1,7 @@
 #ifndef PS2RECOMP_ELF_PARSER_H
 #define PS2RECOMP_ELF_PARSER_H
 
+#include <cstdint>
 #include <elfio/elfio.hpp>
 #include <string>
 #include <vector>


ran-j:
Can you remove the comments the pr description is enough

smmathews:
Done — comment removed, the include is now the whole change.

ran-j:
Thanks we can do like that we can discuss on the pr and change infos and put less comments on codem

PR #209, opened 2026-08-17
Don't promote a whole function because of an indirect call

Sinan-Karakaya:
When a computed jump cannot be resolved to a jump table, every instruction in the function becomes an entry point, because the jump could land on any of them. That fallback was also being applied to JALR.

JALR is not a jump but a call: it transfers control to another function and returns to the instruction after the delay slot. That return address is already queued as a resume target a few lines above, so nothing else in the function needs to be reachable from outside.

Indirect calls are ordinary code — function pointers, virtual dispatch, callbacks — so the fallback fires constantly on real games. Classifying every warned site on Dragon Quest VIII (NTSC-U) by decoding the instruction out of the ELF:

| kind | sites |
|---|---|
| JALR (indirect call) | 2,210 |
| JR (computed jump) | 212 |

with `jalr $t9` alone accounting for 1,874. Split by whether the function contains a genuine computed jump at all:

| | functions | promoted entries |
|---|---|---|
| only indirect calls | 1,078 | 160,100 (84.3%) |
| at least one computed jump | 204 | 29,776 |

This restricts the fallback to JR. Measured on the same binary:

| metric | before | after |
|---|---|---|
| promoted fallback entries | 189,876 | 1,688 |
| registered table entries | 156,783 | 75,386 |
| `register_functions.cpp` | 13 MB | 6.0 MB |
| total generated output | 180 MB | 163 MB |
| largest function file | 1.2 MB | 756 KB |

I checked this did not silently drop anything needed: all 2,210 indirect-call sites were re-examined for their return-address resume entry (`addr + 8`) in the regenerated output. 42 lack one, and all 42 are inside a region of interleaved rodata where the "function" is a carving from string and float data and the return address falls outside its bounds. No real code lost an entry point.

All 11,492 generated translation units still pass `clang++ -std=c++20 -fsyntax-only`.

### DIFF
--- a/ps2xRecomp/src/lib/control_flow_analyzer.cpp
+++ b/ps2xRecomp/src/lib/control_flow_analyzer.cpp
@@ -385,7 +385,14 @@ namespace ps2recomp
                         }
                     }
                 }
-                if (!foundTable)
+                // Only an unresolved computed *jump* can land on an arbitrary
+                // instruction of this function and therefore force every address to
+                // become an entry point. JALR is a call: it transfers control to
+                // another function and comes back to the instruction after the delay
+                // slot, which is already queued as a resume target above. Treating a
+                // call like a jump here promotes the whole function for what is
+                // usually just a function pointer or virtual dispatch.
+                if (!foundTable && jrInst->function != SPECIAL_JALR)
                 {
                     needsIndirectFallback = true;
                 }
--- a/ps2xTest/src/code_generator_tests.cpp
+++ b/ps2xTest/src/code_generator_tests.cpp
@@ -529,7 +529,7 @@ void register_code_generator_tests()
                   "unresolved JR should not pretend it has a resolved local jump table");
     });
 
-    tc.Run("unresolved JALR marks internal labels as indirect fallback resume entries", [](TestCase &t) {
+    tc.Run("unresolved JALR resumes after the call without promoting the function", [](TestCase &t) {
         Function func;
         func.name = "unresolved_jalr_fallback";
         func.start = 0x3200;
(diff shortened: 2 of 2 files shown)


Sinan-Karakaya:
The first CI run failed one test, `unresolved JALR marks internal labels as indirect fallback resume entries`, and I have updated it in the pushed revision. Flagging that explicitly since changing a test alongside the code it covers deserves justification.

The test asserted that `0x320C` lands in `indirectFallbackEntryPoints` for a JALR at `0x3204`. That address is the call's return pc (`0x3204 + 8`), and it is still registered: `queueResumeEntryTarget` puts it in `resumeEntryPoints`, and `ps2_recompiler.cpp` merges both sets into the same per-owner resume-target list:

```cpp
ownerTargets.insert(ownerTargets.end(),
                    analysisResult.resumeEntryPoints.begin(),
                    analysisResult.resumeEntryPoints.end());
ownerTargets.insert(ownerTargets.end(),
                    analysisResult.indirectFallbackEntryPoints.begin(),
                    analysisResult.indirectFallbackEntryPoints.end());
```

So the return pc keeps identical downstream treatment; what changes is that the other instructions in the function are no longer promoted with it. The preceding test, `JALR should mark its return/fallthrough pc as resumable`, already covers the part that matters and still passes untouched.

The updated test now asserts the behaviour rather than the mechanism: the return pc is resumable and emits a label, and unrelated instructions (`0x3210`, and the function start) are *not* promoted — which makes it a regression test for the actual fix. The sibling `unresolved JR ...` test is deliberately left alone, since JR must still promote; I verified locally that it continues to pass.

Empirically, on the binary in the description, all 2,210 indirect-call sites keep their return-address resume entry in the regenerated output. The only 42 without one sit inside a region of interleaved rodata where the "function" is a carving from string data and the return address falls outside its bounds.

PR #234, opened 2026-08-29
Bind the EE scheduler executor thread on first IRQ dispatch

hedgeg0d:
The EE scheduler requires every scheduler operation to run on the executor thread, which gets bound in reset() or bindMainContextForSyscall(). The SIF DMA path (sceSifSetDma -> dispatchDmacHandlersForCause -> dispatchIrq) never binds it, so any direct call from outside the scheduler run loop trips the executor assertion in Debug builds.

Production is unaffected because guest code always runs inside the scheduler loop, and Release builds compile the assertion out, which is why CI never caught it.

dispatchIrq now binds the executor thread on first use, mirroring the lazy binding already done in bindMainContextForSyscall(). The assertion stays in place for every subsequent call.

Verified with the Debug test suite: previously crashed in PS2SifDma after 283 tests, now all 425 tests pass.

### DIFF
--- a/ps2xTest/CMakeLists.txt
+++ b/ps2xTest/CMakeLists.txt
@@ -69,6 +69,11 @@ target_include_directories(ps2_test_lib PRIVATE
     ${CMAKE_SOURCE_DIR}/ps2xRuntime/src/lib
 )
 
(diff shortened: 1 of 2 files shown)


hedgeg0d:
Rebased onto upstream main so the test fix from #233 stays out of this PR.

PR #236, opened 2026-08-29
Clamp same-start functions to Ghidra map boundaries

hedgeg0d:
When a Ghidra CSV and the analyzer heuristics both produce a function at the same start address, the analyzer's wider span wins. Both names are auto-generated on stripped retail binaries (sub_ vs FUN_), so the merge comparator falls through to the larger end, and the authoritative-boundary pass skips auto names entirely. The result is that Ghidra's accurate boundaries never override the heuristic spans, and data blobs inside those spans get decoded as instructions.

Measured on Katamari Damacy (SLUS_210.08): 893k spurious decode errors with the Ghidra map loaded, 0 after this fix.

The change records the CSV end for every map start and uses it as the authoritative end, regardless of the auto-generated name. The analyzer fallback-only starts are still removed when they are not in the map.

Added a regression test that builds a minimal ELF where the JAL-fallback scan widens a function span, then checks the Ghidra map end clamps it.


hedgeg0d:
This duplicates #206, which fixes the same same-start carve vs map-row tie-break with a cleaner approach (dropping all auto carvings and appending the map rows). Closing mine in favor of #206.

PR #238, opened 2026-08-30
fix(runtime): contain translated PS2 paths

hedgeg0d:
PS2 device paths were normalized but not checked against their configured roots. Parent traversal and pre-existing symlinks could resolve outside `hostRoot`, `cdRoot`, or `mcRoot`.

- Canonicalize configured roots and translated paths before accepting them
- Compare existing roots by filesystem identity
- Restrict direct Windows paths to `cdRoot`
- Add translation and `fioOpen` regression tests

Tested with `ps2x_tests`: 428 passed. The runtime changes from #235 were applied only for the Debug test run because current `main` aborts in `PS2SifDma` without that fix.


hedgeg0d:
Replaced by #239. This branch included an unrelated commit from the fork's main branch.

PR #245, opened 2026-09-02
Use occupancy masks for VU pipeline queues

GTTeancum:
## Summary

- Track occupied VU flag, store, VF, VI, and ACC pipeline slots with compact bitmasks.
- Allocate the first free slot with `std::countr_zero`.
- Visit only occupied slots when committing results or checking pending work.

## Motivation

The fixed pipeline arrays are usually sparse. Earlier X-Men Legends bring-up
profiling identified VU execution and pipeline retirement as major costs, so
checking empty slots on every cycle is avoidable work. Allocation and retirement
must retain their original ascending slot order, timing, and reset behavior.

This is not a claim of current playable gameplay speed. Earlier title-screen
timings are not a sustained gameplay benchmark. The September 4 follow-up adds
regression coverage only; it makes no further runtime or performance change.

## Testing

- Windows 11, Visual Studio 2022 Release, one BelowNormal build worker.
- Original branch: 425 passed, 0 failed.
- With the new regression tests: 427 passed, 0 failed.
- Forty independent upper/lower instruction pairs, checked at every cycle
  across 48 one-cycle slices with intervening zero-cycle resumes. Reused VF
  destinations receive alternating values so missing or late later writes are
  observable. The test also checks same-pair FSSET priority and MAC/STATUS.
- Twelve restart/reset cases ensure cancelled pending flag/VF writes cannot
  reappear after a fresh execution, including restarts at nonzero cycle counts.

All new test programs and inputs are synthetic; no game data is included.

### DIFF
--- a/ps2xRuntime/include/runtime/ps2_vu1.h
+++ b/ps2xRuntime/include/runtime/ps2_vu1.h
@@ -216,6 +216,11 @@ private:
     std::array<PendingVfWrite, kMaxPendingVfWrites> m_vfWritePipeline{};
     std::array<PendingViWrite, kMaxPendingViWrites> m_viWritePipeline{};
     std::array<PendingAccWrite, kMaxPendingAccWrites> m_accWritePipeline{};
+    uint32_t m_flagPipelineMask = 0;
+    uint32_t m_storePipelineMask = 0;
+    uint32_t m_vfWritePipelineMask = 0;
+    uint32_t m_viWritePipelineMask = 0;
+    uint32_t m_accWritePipelineMask = 0;
     XgkickPipeline m_xgkick{};
     std::array<std::array<uint64_t, 4>, 32> m_vfReady{};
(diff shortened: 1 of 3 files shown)


GTTeancum:
Added a small follow-up to clear retired FlagPipelineEntry objects directly. With MSVC 19.44, the aggregate assignment built a zeroed stack temporary and copied it back; direct clearing emits zero stores to the integer/bool entry instead. Deadline checks, slot retirement order, sticky accumulation, and occupancy masks are unchanged. The existing upstream-based Release suite passes 427/427. In the separate X-Men development branch, the analogous interpreter/native-block change also passes 126 focused VU tests and exact recorded state, memory, GIF bytes and cycle checks at normal and 1/8/16/64-cycle slicing. Those game-specific replay fixtures and native-block changes are not part of this PR. No whole-game FPS claim is made.

